PR #46, opened 2021-02-24
176936761 improve software version score

tvannurden:
#### What's this PR do?
- Grants a score of 2 if validator is running an edge version of the validator software
- Refactors, creating `ValidatorSoftwareVersion` to make comparisons easier to read


brianlong:
@tvannurden Will the tests fail if we change `config/cluster.yml` in the future? Possibly use separate files for each environment: `config/cluster.test.yml`, `config/cluster.development.yml`, `config/cluster.production.yml`. I'm open to other ideas too.

brianlong:
We could also create sections for each environment within a single `config/cluster.yml` file.

brianlong:
Nice idea to use `Gem::Version`!

brianlong:
I would like to see some tests on the patch level too. If the reference version is `1.4.8`, then version `1.4.7` will be one point. I see that you have a method for `running_latest_minor?`. Let's add another method called `running_latest_minor_and_patch?` and then create some tests around that.

brianlong:
FWIW, if the major version is less than the reference, then the score is always 0. A newer major version is OK.

tvannurden:
> @tvannurden Will the tests fail if we change `config/cluster.yml` in the future? Possibly use separate files for each environment: `config/cluster.test.yml`, `config/cluster.development.yml`, `config/cluster.production.yml`. I'm open to other ideas too.

Good point - They will eventually fail, yes. What I can do as a simple workaround is stub the current version for the sake of the tests. I can also demonstrate some different scenarios to prove the code will still work far into the future. For example when the version is `1.5.99`, or `1.9.` or `12.5.3`, etc. I don't think it makes sense to marry our `Rails.env` with the Solana cluster version, like we already have with mainnet and testnet.

brianlong:
@tvannurden This looks good, but I just thought of something -- I don't want to put the RPC pool IPs for production in the public Git repo. The current version of `config/cluster.yml` is OK for development & testing. Let's re-configure this so that we can symlink the production file from the shared folder on the server. We'll need to update the linked files in the Capistrano config too. .gitignore as well.

tvannurden:
Sounds good @brianlong. I have a good idea for what to do. I'll create a ticket and jot my ideas down there.

brianlong:
That is a blocker for this PR. Let me know when it is ready and I'll merge it ASAP.

tvannurden:
All set @brianlong. I've added `/home/deploy/validators.app/shared/config/cluster.yml` to the server and Capistrano is setup to symlink to `/home/deploy/validators.app/current/config/cluster.yml` on deploy. I've also added `cluster.yml` to `.gitignore`.

PR #54, opened 2021-03-09
177161884 validator block histories n+1

tvannurden:
#### What's this PR do?
- Fix the N+1 query described in the [PT Ticket](https://www.pivotaltracker.com/story/show/177161884)


test/logic/validator_score_v1_logic_test.rb
```diff
@@ -96,6 +96,25 @@ def setup
   end
 
   test 'block_history_get' do
+    Timecop.scale do
```
tvannurden, line 99: `Timecop.scale` makes milliseconds seem like minutes to ensure these get different `created_at` timestamps

tvannurden: https://github.com/travisjeffery/timecop#timecopscale

app/logic/validator_score_v1_logic.rb
```diff
@@ -209,14 +208,27 @@ def block_history_get
           p.payload[:batch_uuid]
         )
 
-      # TODO: Eliminate the N+1 Query caused by
-      # validator.validator_block_histories.last
+      validator_ids = p.payload[:validators].map { |v| v.id }
+
+      sql = <<-SQL_END
```
tvannurden, line 213: @brianlong LMK if you can think of a more elegant solution than what I came up with.

brianlong:
@tvannurden Please spend a little more time trying to optimize the query. I ran a trial deployment from this branch and the query time was significantly longer than our current solution. Feel free to deploy this to the stage server so you can experiment with a bigger dataset. When you are done on the staging server, please re-deploy with the `new-ui` branch so we can see @kbiala's latest work.

app/logic/validator_score_v1_logic.rb
brianlong: Try using the :network & :batch_uuid values to grab only the current validator_block_histories. We should have a DB index for [:network, :batch_uuid].

brianlong:
@tvannurden Is it easy to un-merge `new-ui`? I don't think that branch is ready for production yet.

kbiala:
You can create temporary 'stage' branch and merge both into it, that should be much easier that reverting my changes.

tvannurden:
> @tvannurden Is it easy to un-merge `new-ui`? I don't think that branch is ready for production yet.

It is - do you want me to unmerge it @brianlong ? I had only merged it in temporarily in order to get it deployed to stage while keeping Klaudia's UI changes

brianlong:
@tvannurden Yes, before we merge into master. FWIW, it is OK If the staging server jumps around between branches. We don't have real users there. Ready for me to review?

tvannurden:
I reverted the code in `new-ui` so ready to review again @brianlong

tvannurden:
This is looking good in stage @brianlong - looks like we're saving about 4s on the background job. Feel free to merge whenever comfortable.

brianlong:
@tvannurden On my system, the tests were green and now they are not -- not sure why. Here is the error:
```
Error:
ValidatorScoreV1LogicTest#test_block_history_get:
NameError: uninitialized constant ValidatorScoreV1LogicTest::Timecop
    test/logic/validator_score_v1_logic_test.rb:114:in `block in <class:ValidatorScoreV1LogicTest>'
```
Do you have any insights here?

tvannurden:
`bundle install` @brianlong

brianlong:
> `bundle install` @brianlong

Tried that!

brianlong:
```
rm Gemfile.lock
bundle install
rails test => green!
```

tvannurden:
That's ironically what I just tried as well and it worked for me too! -> I have no idea why though 🤔

brianlong:
@tvannurden I just bumped Rails & gem versions on the master branch. The deployment was clean. Please merge master into whichever branch you are working on.

PR #59, opened 2021-03-13
177163675 skipped slots moving avg

tvannurden:
#### What's this PR do?
Adds a new graph-able metric, skipped slot % moving average, to `ValidatorBlockHistoryStat` and `ValidatorBlockHistory`.
`skipped_slot_percent_moving_average` is a static, persisted attribute that holds the 24h moving average of either a validator (`ValidatorBlockHistory`) or the cluster's (`ValidatorBlockHistoryStat`).

#### Callback
When new `ValidatorBlockHistory` and `ValidatorBlockHistoryStat` records are created, calculate the moving average for the previous 24h. Calculating this on create makes viewing the chart performant.

#### Rake task to populate stats for existing records
In order to populate `skipped_slot_percent_moving_average` for existing records, two rake tasks have been added. 

`rails db: generate_vbh_skipped_slot_percent_moving_averages` only calculates `skipped_slot_percent_moving_average` for records created within the last day because the front-end doesn't need `skipped_slot_percent_moving_average` for older records. Technically we'd just need the last 240 (about last 8-hours) from each Validator but I ended up just rounding up to 1-day. In production, this rake task should take n-minutes to run.

`rails db: generate_vbhs_skipped_slot_percent_moving_averages` only calculates for last 7-days for a similar reason. In production, this rake task should take x-minutes to run.

This is difficult to test end-to-end locally without pulling down stage or production data. Instead, I plan on deploying and testing through stage. My local testing has been done by pulling down stage data and running off that. Note: I had to install and am running a MySQL 8.0 Docker Container in order to do this. @brianlong if you want me to show you how I did this, although it's fairly simple.

https://www.pivotaltracker.com/story/show/177163675

#### Things to watch out for


app/models/validator_score_v1.rb
```diff
@@ -64,6 +64,13 @@ class ValidatorScoreV1 < ApplicationRecord
   serialize :skipped_slot_history, JSON
   serialize :skipped_after_history, JSON
 
+  def skipped_slot_history_moving_averages
+    ValidatorBlockHistory
+      .where(validator: validator, network: network)
+      .last(60)
```
tvannurden, line 70: [The front-end only looks at the last 60 records](https://github.com/brianlong/www.validators.app/blob/master/app/views/validators/index.html.erb#L242-L247) so we only need to load 60.

app/views/validators/index.html.erb
```diff
@@ -304,13 +304,25 @@
...
-                 %>
+                    slot_vector = validator.score&.skipped_slot_history.to_a[-vl..-1].map { |v| (v.to_f*100.0).round(1) }
+
+                    moving_avg_color = if validator.skipped_slot_score == 2
+                        '#7fbf7f'
+                      elsif validator.skipped_slot_score == 1
+                        '#7f7fff'
+                      else
+                        '#bfbfbf'
+                      end
```
tvannurden, line 315: These colors are just darker shades of the existing colors. @kbiala feel free to make any changes you recommend whenever you get to this.

app/models/validator_block_history.rb
```diff
@@ -45,4 +51,16 @@ def self.median_skipped_slot_percent_for(network, batch_uuid)
       batch_uuid: batch_uuid
     ).median(:skipped_slot_percent)
```
brianlong, line 52: Let's calculate the median moving average instead of :skipped_slot_percent. Same for self.average_skipped_slot_percent_for. We will calculate the scores from the MA, not the instant values. You'll need to chase this through `app/logic/validator_score_v1_logic.rb` too.

app/views/validators/_block_history_chart.html.erb
```diff
@@ -9,31 +9,50 @@
...
+        {
+          label: 'Skip %',
+          fill: false,
+          backgroundColor: 'Green',
+          borderColor: 'Green',
+          borderWidth: 1,
+          radius: 0,
+          data: <%= @data.map { |k,v| v[:skipped_slot_percent].round(1) }%>
+        },
+        {
```
brianlong, line 27: Maybe we can drop this line from the chart -- just show the 'Cluster Moving Avg Skip %'

app/views/validators/_block_history_chart.html.erb
```diff
@@ -9,31 +9,50 @@
...
-          }
-        ]
+      // x-axis labels
+      labels: <%= @data.keys %>,
+      // An Array of JSON hashes for each dataset to plot. The y-axis is
+      // dynamicly calculated. Look for othe options to control the y-axis if
+      // desired.
+      datasets: [
+        {
+          label: 'Skip %',
```
brianlong, line 19: Dotted line in a lighter color?

kbiala:
@tvannurden Is this PR ready to merge to my branch? I'd like to play with the colors&design to see what works best?

app/views/validators/_block_history_chart.html.erb
tvannurden: Sure, I'll just comment out for now so that we can add back later if wanted.

app/views/validators/_block_history_chart.html.erb
tvannurden: Here's what it looks like w/ that change @brianlong 

image

brianlong: Looks good

tvannurden:
@brianlong this is ready for another review. The only thing I have left to do is to finish the rake task to populate `validator_score_v1s.skipped_slot_moving_average_history` for the existing records - otherwise, the graph on the home page will be incomplete until 8-hours have elapsed after deploy.

PR #60, opened 2021-03-18
177372602 new script for updating avatars

maciejkocylapc:
#### What's this PR do?
script validators_update_avatar_url.rb to check for dead urls and replace them if newer exists

from main directory rails r script/validators_update_avatar_url.rb

https://www.pivotaltracker.com/story/show/177372602

#### Things to watch out for


config/credentials.yml.enc
```diff
@@ -1 +1 @@
...
...GCYF7dBk1jDAZeuZ3HopHYbxZ7HFIck2IJVLPtwL3y5eYhfYZPE6knAN432jzp7cnlJQ/yBQGivCawcWqGTwGfi23flDlEJO6Kx5y2f+9JFbqbIGShpzcvTcP56n+AC2U+XeCf/YfWbuacfJQ0u48SfylWAIN/qntd9zfBBzPL+3tYTdd5eX1z/6OZGfDYIlBTnjDHGbiLf6eGni3fvCOn7R/6rrI4GSuJY5BkuM1iCvX5grGj1nJSfcNiz79EX0DlgrQ8gD3mA6pHfq8WyddBd9ue3N+MJTK0DvjnRf6q6x+HILOUF3TP5yhfK7uO1Nx0hBYzcMDoV6HWmOaxF3AI62ipRccNFTDF00cXcZDZYBNf+lths+57DUufB618hg1XXpm14HZsprnuvH+7/qlZ5pmZUv5nGIp/w+KIViUSLpgvkdNa3vywTttILvLMaQoo95brnX/vHy8LiOUmx5Dhw3b5VNy/pJBHi/YPK/mzCvbVrOT6UIMgwWgOwnj30xw5B9gYvd3VZG4QXkSJL7pTN5qp2vioHE7Q7zRTERxxZ61G8WFjFhANJgYn6Ma6zrBx6MwN86tyJ0q8RSMacqI+pPXQ0HXMeOK+pwvr6mo7GlD5zeWpddFbJsncXO9kjfA8YQI5ZkejVnRdBKOejBTHK1CeP9AO8J7z7YXGiogofp6jbrdecykNTqXYlSwQuHo+EUcQ8aMuiGojXRZtVW+2DNoeR3Ms5lzuTU+kKj/veLNHqTPmIg79Jv0FSO9ylz+UkuAzeoJlfS3KwJyhwTMR4IJSVqOgwe8C+xEe0ywP4B0O0+txBUsg0yQOzOlKph2ZQ/zJqpSF0UMgv+3UJsAdIZ/tYrkoEt9JqqQwEU8LfSrwE/EIaFAsTHPbEhZJZZYllsxSzLoHxVAJQk6MCwLWsRXHkfvg01MT1BfAGUnv6crjvcLonA+uSoHxijwCbxMdtokoYRHx6/vIgOFywWiRTMgXrOKfB1jJyPIa0HF7Zk+3B04RtGY+GwJhugMxYAOg3b4VqCE/IG2rEQ9Qf/CwuO1PAHWVpSgp0Hk2TqLcx7nm2DZ7rJVrJX6xacWfGmjhrPkI1fzfW6gS9R90dDaQqptrqjMeOpI30zUYHxuMt1reM7b603m36zGlXJfivH4D80--c6lnZYpOeKkI4IeO--F9Y88xORPJx7BUffPAfaog==
```
brianlong, line 1: Did anything actually change here? If not, please revert this file and get the development key from @kbiala.

app/services/keybase_service.rb
```diff
@@ -0,0 +1,27 @@
...
+
+      if Nokogiri::HTML(keybase[:html]).css(image_css).present?
+        avatar_url = Nokogiri::HTML(keybase[:html]).css(image_css)
+                            .attribute('src').value
+        if avatar_url.to_s.include?('keybase_processed_uploads')
+          puts "update #{validator.name} avatar to #{avatar_url}"
+          validator.update(avatar_url: avatar_url)
+        end
+      end
+    end
```
brianlong, line 24: If the call to Keybase fails, let's update the validator record with the default avatar URL at 'https://keybase.io/images/no-photo/placeholder-avatar-180-x-180@2x.png'

app/logic/agent_logic.rb
```diff
@@ -79,7 +79,8 @@ def get_page(params = {})
           result[:errors] << e.message + (proxy.present? ? " -- #{proxy[:host]}:#{proxy[:port]}" : '')
           add_error(params[:url], result[:http_code], e, proxy_for_log, result[:tries])
           sleep(sleep_secs) unless Rails.env == 'test'
-        rescue Net::HTTPServerException => e
+        rescue Net::HTTPClientException => e #Net::HTTPServerException was deprecated
```
tvannurden, line 82:
```suggestion
        rescue Net::HTTPClientException => e
```

app/logic/agent_logic.rb
```diff
@@ -79,7 +79,8 @@ def get_page(params = {})
...
+          puts e
```
tvannurden, line 83: Can remove

app/services/keybase_service.rb
```diff
@@ -0,0 +1,27 @@
...
+  def get_validator_avatar(validator)
+    image_css = '.picture > .img-circle'
+    keybase_url = @keybase_url + validator.keybase_id
+    keybase = get_page(url: keybase_url)
+
+    if keybase[:http_code] == '200'
+
+      if Nokogiri::HTML(keybase[:html]).css(image_css).present?
+        avatar_url = Nokogiri::HTML(keybase[:html]).css(image_css)
+                            .attribute('src').value
```
tvannurden, line 18: If you're going to line break, line break all the method calls
```suggestion
        avatar_url = Nokogiri::HTML(keybase[:html])
                               .css(image_css)
                               .attribute('src').value
```

app/services/keybase_service.rb
```diff
@@ -0,0 +1,27 @@
...
+        if avatar_url.to_s.include?('keybase_processed_uploads')
+          puts "update #{validator.name} avatar to #{avatar_url}"
```
tvannurden, line 20: What's this intended for?

app/services/keybase_service.rb
```diff
@@ -0,0 +1,27 @@
+class KeyBaseService
```
tvannurden, line 1: The class should be called `KeybaseService` if the filename is `keybase_service.rb`. Might want to double check what you want here. Is it `Key Base` or `Keybase`?

script/validators_update_avatar_url.rb
```diff
@@ -0,0 +1,29 @@
...
+interrupted = false
+trap('INT') { interrupted = true }
+
+agent = Mechanize.new
+keybase_service = KeyBaseService.new
+
+%w[testnet mainnet].each do |network|
+  Validator.where(
+    ["network = ? and keybase_id != '' AND avatar_url IS NOT NULL", network]
+  ).each do |validator|
```
tvannurden, line 20: Use Active Record for simple queries like this. It's easier to read than just a raw query.

script/validators_update_avatar_url.rb
```diff
@@ -0,0 +1,29 @@
...
+    begin
+      validator_avatar = agent.get(validator.avatar_url)
+    rescue StandardError => e
+      if e.response_code && ['404', '403'].include?(e.response_code)
+        keybase_service.get_validator_avatar(validator)
+      end
+    end
```
tvannurden, line 27: I don't think this code is doing anything. Are you sure this works?

script/validators_update_avatar_url.rb
```diff
@@ -0,0 +1,29 @@
...
+require File.expand_path('./app/services/keybase_service.rb')
+# Default URL is at 'https://keybase.io/images/no-photo/placeholder-avatar-180-x-180@2x.png'
+
+# include AgentLogic
+# include KeyBaseService
+
+interrupted = false
+trap('INT') { interrupted = true }
+
+agent = Mechanize.new
```
tvannurden, line 14: Might be overkill to use Mechanize here. `Net::HTTP` or `RestClient` might be better.

app/services/keybase_service.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  include AgentLogic
+
+  def initialize
+    @keybase_url = "https://keybase.io/"
+  end
+
+  def get_validator_avatar(validator)
```
tvannurden, line 9: I'd recommend changing this method name. When I see a method that has `get` in it, I'd think it's **only** retrieving something. Whereas your method gets _and_ sets `Validator#url`

script/validators_update_avatar_url.rb
```diff
@@ -0,0 +1,29 @@
...
+keybase_service = KeyBaseService.new
+
+%w[testnet mainnet].each do |network|
+  Validator.where(
+    ["network = ? and keybase_id != '' AND avatar_url IS NOT NULL", network]
+  ).each do |validator|
+    begin
+      validator_avatar = agent.get(validator.avatar_url)
```
tvannurden, line 22: You're assigning `validator_avatar` but doing nothing with it.

brianlong:
@maciejkocylapc My goal for the script `script/validators_update_avatar_url.rb` is to update all validators one-time per day. As it stands, the script will only update new validators (remove "AND avatar_url IS NOT NULL").

We need a test for the keybase fetch. Tyler is correct in that we shouldn't try to do perform multiple actions in the same method there. A simpler implementation of the Keybase service might be to express Keybase as a module with a simple function inside:
```
module Keybase
  # Example use:
  # get_avatar_url(keybase_id) => 'https://path/to/avatar.png'

  def get_avatar_url(keybase_id)
    # Use the keybase_id to fetch the page and return the avatar URL, or the default URL if there is a problem, as a string.
  end
end
```
That module function can be tested easily and is thread-safe if needed.

The process of fetching the URL and then saving the validator can be simply handled within the script. We don't really need the Service object.

Once we have the keybase module done, we can modify the other script to use the same function.

app/services/keybase_service.rb
maciejkocylapc: I'm afraid that if I do that, we'll never catch the future updates of this url. maybe I should clear it instead?

config/credentials.yml.enc
maciejkocylapc: I'm not sure if there should be ports after solana testnet urls? there's separate variable rpc_port and for me it worked when I removed the ports in links.

brianlong: Your tests should pass with the current credentials file -- no changes required for this story. Do you also have `config/cluster.yml`?

app/services/keybase_service.rb
brianlong: Hmmm... We're using a null value in the avatar_url column to determine new v. existing validators. The daily update script will pick up changes if the validator later provides a photo to Keybase.

It sounds like you are concerned that Keybase might change the location of their default image and then we will show a bunch of broken image links. We can put a placeholder image on our site and link to the local URL instead when we need a placeholder image. Sound good?

maciejkocylapc: Yes, but should I also check for the newer version of the avatar if validator has default url in our db ? (this one should not be broken) And if so, shouldn't we be worried about requesting keybase too much ?

brianlong: Yes, the daily update script should check Keybase for new URLs for all validators. Replace our default if we see a URL on Keybase. Add a 1-second sleep between requests on the daily script so we don't hammer their servers.

brianlong:
@maciejkocylapc Ready for us to review again?

maciejkocylapc:
> @maciejkocylapc Ready for us to review again?

Yes

PR #63, opened 2021-03-26
skipped votes history added to validator_score_v1

maciejkocylapc:
#### What's this PR do?
Skipped vote % showed as chart instead of number 

root URL should contain chart of skipped votes

https://www.pivotaltracker.com/story/show/177459120

#### Things to watch out for


db/schema.rb
```diff
@@ -33,28 +33,6 @@
     t.index ["key"], name: "index_active_storage_blobs_on_key", unique: true
   end
 
-  create_table "audits", charset: "utf8", force: :cascade do |t|
```
tvannurden, line 36: I had to double check these diffs (for `block_commitments`, `slots`, and `stake_boss_stake_accounts`), but they look good. There are no migration files for these so they don't belong in `schema.rb`. `rails db:migrate` should always produce the same result as `rails db:schema:load`

test/fixtures/vote_account_histories.yml
```diff
@@ -7,6 +7,9 @@ one:
   last_vote: 1
   activated_stake: 1_000_000_000
   credits: 1
+  credits_current: 193916
+  slot_index_current: 299198
+  skipped_vote_percent_moving_average: 0.1
```
tvannurden, line 12: This is ultimately up to @brianlong, but I highly recommend using `FactoryBot` instead of the Rails fixtures. Rails fixtures don't use or call ActiveRecord validations or callbacks, so the objects you create via fixtures can end up looking completely different than what would be permissable or happen in the real app.

tvannurden (commented):
This is looking pretty good for an initial pass on it. I'd like to see it run in stage for a while after we get https://github.com/brianlong/www.validators.app/pull/59 merged and deployed. I know this part hasn't been implemented yet, but actually displaying the new chart should be done in a separate PR (and PT ticket) because the data won't exist when it first gets deployed. You have to wait at least 32-hours (24-hours + 8-hours) in order to have full stats to display.

app/logic/validator_score_v1_logic.rb
```diff
@@ -81,10 +80,18 @@ def block_vote_history_get
         batch_uuid: p.payload[:batch_uuid]
       ).where({ account: p.payload[:validators].map(&:account) }).to_a
 
+      vote_histories = VoteAccountHistory.joins(:vote_account).where(
+        network: p.payload[:network],
+        batch_uuid: p.payload[:batch_uuid]
+      ).where('vote_account.validator_id': p.payload[:validators].pluck(:id)).to_a
```
tvannurden, line 86: Nice! This is probably more performant than the similar query I had for Skipped Slot %

app/logic/validator_score_v1_logic.rb
```diff
@@ -81,10 +80,18 @@ def block_vote_history_get
...
       p.payload[:validators].each do |validator|
         # Get the last root & vote for this validator
         vh = validator_histories.select { |vh| vh.account == validator.account }.first
-
+        vote_h = vote_histories.select { |vote_h| vote_h.vote_account.validator_id == validator.id }.first
```
tvannurden, line 90:
```suggestion
        vote_h = vote_histories.find { |vote_h| vote_h.vote_account.validator_id == validator.id }
```

app/models/validator_score_v1.rb
```diff
@@ -151,6 +155,28 @@ def root_distance_history_push(val)
     end
   end
 
+  def skipped_vote_percent_moving_average_history_push(val)
+    self.skipped_vote_percent_moving_average_history = [] if skipped_vote_percent_moving_average_history.nil?
+
+    skipped_vote_percent_moving_average_history << val
+
+    # Prune the array  to include the most recent values
```
tvannurden, line 163: I think this comment is redundant by now. Please remove

app/models/validator_score_v1.rb
```diff
@@ -151,6 +155,28 @@ def root_distance_history_push(val)
...
+      self.skipped_vote_percent_moving_average_history = skipped_vote_percent_moving_average_history[-MAX_HISTORY..-1]
+    end
+  end
+
+  def skipped_vote_history_push(val)
+    self.skipped_vote_history = [] if skipped_vote_history.nil?
+
+    skipped_vote_history << val
+
+    # Prune the array  to include the most recent values
```
tvannurden, line 174: I think this comment is redundant by now. Please remove

app/models/vote_account_history.rb
```diff
@@ -34,4 +35,52 @@
...
+
+  def self.median_skipped_vote_percent_for(network:, batch_uuid:)
+    vah_skipped = where(
+      network: network,
+      batch_uuid: batch_uuid
+    ).map(&:skipped_vote_percent)
+    vah_skipped.sort[vah_skipped.length/2]
+  end
+
+  def self.average_skipped_vote_percent_moving_average_for(network:, batch_uuid:)
```
tvannurden, line 73: This isn't used/called from anywhere. What's it intended for? Please leave a comment for what it's going to be used for in the future if it really is needed.

app/models/vote_account_history.rb
```diff
@@ -34,4 +35,52 @@
...
+    where(
+      network: network,
+      batch_uuid: batch_uuid
+    ).average(:skipped_vote_percent_moving_average)
+  end
+
+  def self.median_skipped_vote_percent_moving_average_for(network:, batch_uuid:)
```
tvannurden, line 80: This isn't used/called from anywhere. What's it intended for? Please leave a comment for what it's going to be used for in the future if it really is needed.

tvannurden (commented):
Looking better @maciejkocylapc but a few more small things. I think after these revisions we can deploy it to stage and see how it looks there.

tvannurden:
Removing the WIP and "Draft" on this PR since I've already reviewed it.

app/models/vote_account_history.rb
maciejkocylapc: I'm not sure which one of these (average_skipped_vote_percent_moving_average_for or average_skipped_vote_percent_for) should be used in charts - if you're certain that this one is not needed I can remove it.

PR #79, opened 2021-04-20
177633346 Cluster median and average distances

maciejkocylapc:
#### What's this PR do?
show median and average distances of blocks and votes on main page

there should be correct values of median and average for root distance and vote distance over the validator index

https://www.pivotaltracker.com/story/show/177633346

#### Things to watch out for


app/logic/validator_score_v1_logic.rb
```diff
@@ -142,6 +142,13 @@ def assign_block_and_vote_scores
       vote_distance_all_average = array_average(vote_distance_all)
       vote_distance_all_median = array_median(vote_distance_all)
 
+      p.payload[:this_batch].update(
```
tvannurden, line 145: `update!` probably makes more sense here because you aren't specifically checking if the update succeeded or not. Otherwise we don't know if the update actually succeeded or not whereas `update!` will raise an exception and an error bubble up which I'll find out about right away.

tvannurden (changes requested):
Looks good! I just had one comment on this one.

PR #80, opened 2021-04-22
177580613 guard_stake_account method

maciejkocylapc:
#### What's this PR do?
enchance guard_stake_account method

only automatic tests for now

https://www.pivotaltracker.com/story/show/177580613

#### Things to watch out for


kbiala:
@maciejkocylapc Please resolve conflicts

app/models/solana/stake_account.rb
```diff
@@ -60,6 +60,7 @@ def initialize(args)
     # Gets the output from the Solana CLI and populates the attributes
     def get
       ssa = cli_request("stake-account #{@address}", @rpc_urls)
+      @cli_error = ssa[:cli_error]
```
tvannurden, line 63: Interesting...How come you're accessing this key as a symbol but all the others as a string? Some consistency would be nice.

app/logic/solana_logic.rb
```diff
@@ -495,8 +497,12 @@ def cli_request(cli_method, rpc_urls)
       end
 
       # byebug
-
-      return JSON.parse(response_utf8) unless response_utf8 == ''
+      if response_utf8 != ''
+        # Include errors from the solana-cli 
+        return JSON.parse(response_utf8).merge!('cli_error' => error_msg)
```
tvannurden, line 502: Remember that Ruby has implicit return for whatever is evaluated last, so the `return` isn't needed and can be removed.

tvannurden, line 502: `#merge` would be better to use here instead of `#merge!`

tvannurden, line 502: I think you could refactor the conditions in your `if/elseif` statement. It's usually a code smell to see both statements with a `!` in them. I would recommend using `#present?` and `#blank?`

app/logic/solana_logic.rb
```diff
@@ -495,8 +497,12 @@ def cli_request(cli_method, rpc_urls)
...
+      elsif !error_msg.blank?
+        return {cli_error: error_msg}
```
tvannurden, line 504: In one scenario you return a JSON string, in the other you return a Hash. Please choose one and make them consistent.

tvannurden (commented):
@maciejkocylapc here's some comments for a first pass on the review.

tvannurden:
Some other comments for you @maciejkocylapc:

- in `solana_cli_service.rb`, returning an Array if `full_resp = true` and a single value if it isn't passed is very hacky. Please either refactor other areas of the code that call `::request`, or suggest an alternate solution
- In `solana_logic.rb`, `exit_status` is unused
- In `solana_logic.rb#cli_request`: why return an empty array if no data returned? Something else, like `nil` would be be better
- In `solana_logic.rb#cli_request`, instead of using `rpc_urls.each do`, try to use `rpc_urls.find do`

tvannurden:
@maciejkocylapc any updates?

PR #96, opened 2021-05-12
178085324 fix skipped vote

maciejkocylapc:
#### What's this PR do?
- fix skipped vote % for validator index

One of the validators on index page should have 0.00% skipped vote percent (Leader)

https://www.pivotaltracker.com/story/show/178085324

#### Things to watch out for


app/models/vote_account.rb
```diff
@@ -32,6 +32,10 @@ def vote_account_history_last
     vote_account_histories.last
   end
 
+  def vote_account_history_last_in_batch(batch_uuid)
+    vote_account_histories.where(batch_uuid: batch_uuid).last
```
brianlong, line 36: There should be only one vote_account_history for the given batch. I think you can remove `.last` since that will only slow down the results.

brianlong: Rename to `vote_account_history_for(batch_uuid)`

kbiala (approved):

PR #110, opened 2021-05-26
search by vote account

maciejkocylapc:
#### What's this PR do?
search of validators will also include vote_account

get any validators vote account address and try to search by it. It should still be possible to search by
other params.

https://www.pivotaltracker.com/story/show/178277497

#### Things to watch out for


app/controllers/public_controller.rb
```diff
@@ -3,29 +3,31 @@
 # PublicController
 class PublicController < ApplicationController
   def index
-    @sort_order = if params[:order] == 'score'
+    @sort_order = if index_params[:order] == 'score'
```
radek-polcode, line 6: case expression fits more here

radek-polcode (changes requested):

radek-polcode (approved):
Works fine, but there is no information about the fields I could use in search, maybe there should be some info in the placeholder like "Search by name, account, data center key..." or sth like that.

PR #119, opened 2021-06-11
178459409 drop name and random search options

kbiala:
#### What's this PR do?
- Removes 'name' and 'random' options from view, keeps the logic in case we want to bring them back in future

- Use sort options on homepage

- https://www.pivotaltracker.com/n/projects/2177859/stories/178459409

#### Things to watch out for


radek-polcode (commented):
Shouldn't we remove these options also from model method? https://github.com/brianlong/www.validators.app/blob/178459409-drop-name-and-random-sort-options/app/models/validator.rb#L46

PR #121, opened 2021-06-12
178318190 skipped vote median and average calculated in backround

maciejkocylapc:
#### What's this PR do?
reduce the load on the index page by moving skipped_vote calculations to validator_score_v1_logic

check if the numbers and charts are correct on the main page

https://www.pivotaltracker.com/story/show/178318190

#### Things to watch out for

### DIFF
--- a/app/helpers/validators_helper.rb
+++ b/app/helpers/validators_helper.rb
@@ -53,7 +53,7 @@ module ValidatorsHelper
   end
 
   def skipped_vote_percent(validator, batch)
-    if validator.score&.skipped_vote_history
+    if validator.score&.skipped_vote_history && batch.best_skipped_vote
       # Get the last skipped_vote data from history
       skipped_votes_percent = validator.score.skipped_vote_history[-1]
 
--- a/app/helpers/validators_helper.rb
+++ b/app/helpers/validators_helper.rb
@@ -54,7 +54,10 @@ module ValidatorsHelper
 
   def skipped_vote_percent(validator, batch)
     if validator.score&.skipped_vote_history
+      # Get the last skipped_vote data from history
       skipped_votes_percent = validator.score.skipped_vote_history[-1]
+
+      # Calculate the distance from the best skipped vote and round
       ((batch.best_skipped_vote - skipped_votes_percent.to_f) * 100.0).round(2)
     else
       nil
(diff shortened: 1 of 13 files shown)


radek-polcode:
undefined method `-' for nil:NilClass

 # Calculate the distance from the best skipped vote and round
      ((batch.best_skipped_vote - skipped_votes_percent.to_f) * 100.0).round(2)

`app/helpers/validators_helper.rb:61:in `skipped_vote_percent'`

app/views/validators/_spark_line_skipped_vote.html.erb
```diff
@@ -5,9 +5,9 @@
         type: 'gauge',
         data: {
             datasets: [{
-                data: [-5, -3.0, 0.05],
+                data: [<%= -(@batch.skipped_vote_all_median * 20) %>, <%= -(@batch.skipped_vote_all_median * 10) %>, 0.05],
```
radek-polcode, line 8: Why we multiplicate by 10 and 20 and by 30 below? Maybe you could create some meaningful method in the model to get rid of calculations in the view?

kbiala: I don't think it should be placed in the model - it does not represent any data other than ranges in frontend chart.

kbiala: Data is multiplicated for chart purposes only - first range is eqal median, second median*2, and so on.

radek-polcode: However, please consider using some variables like - for someone who does not work on that these numbers are confusing.

kbiala: @maciejkocylapc I'd pass them in render helper like account and value, I think that will be clear enough

app/helpers/validators_helper.rb
```diff
@@ -53,7 +53,7 @@ def current_software_version(batch, network)
   end
 
   def skipped_vote_percent(validator, batch)
-    if validator.score&.skipped_vote_history
+    if validator.score&.skipped_vote_history && batch.best_skipped_vote
```
radek-polcode, line 56: You can also use guard clause instead of if-else.

radek-polcode (approved):

PR #145, opened 2021-07-08
 178792573-commission-change-report

maciejkocylapc:
#### What's this PR do?
commission_history should be created every time the commission is changed for the validator

run score and gather scripts, check if commission_histories are being created
and if the data is correct

https://www.pivotaltracker.com/story/show/178792573

#### Things to watch out for

### DIFF
--- a/app/models/validator_history.rb
+++ b/app/models/validator_history.rb
@@ -30,6 +30,21 @@ class ValidatorHistory < ApplicationRecord
   # Use the monkey patch for median
   include PipelineLogic
 
+  has_one :commission_history
+
+  after_create :create_commission_history
+
+  def create_commission_history
+    CreateCommissionHistoryService.new(self).call
+  end
+
+  def previous
+    ValidatorHistory.where(account: account)
+                    .where('created_at < ?', created_at)
+                    .order(created_at: :desc)
+                    .last
+  end
+
   class << self
     def for_batch(network, batch_uuid)
       where(network: network, batch_uuid: batch_uuid)
--- a/app/models/validator_history.rb
+++ b/app/models/validator_history.rb
@@ -30,20 +30,6 @@ class ValidatorHistory < ApplicationRecord
   # Use the monkey patch for median
   include PipelineLogic
 
-  after_create :create_commission_history
-
-  def create_commission_history
-    CreateCommissionHistoryService.new(self).call
-  end
-
-  # single previous record from the same validator
-  def previous
-    ValidatorHistory.where(account: account)
-                    .where('created_at < ?', created_at)
-                    .order(created_at: :desc)
-                    .first
-  end
-
   class << self
     def for_batch(network, batch_uuid)
       where(network: network, batch_uuid: batch_uuid)
--- a/app/models/validator_history.rb
+++ b/app/models/validator_history.rb
@@ -30,19 +30,18 @@ class ValidatorHistory < ApplicationRecord
   # Use the monkey patch for median
   include PipelineLogic
 
-  has_one :commission_history
-
   after_create :create_commission_history
 
   def create_commission_history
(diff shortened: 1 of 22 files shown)


radek-polcode (approved):

app/models/validator_history.rb
```diff
@@ -30,6 +30,20 @@ class ValidatorHistory < ApplicationRecord
   # Use the monkey patch for median
   include PipelineLogic
 
+  after_create :create_commission_history
```
kbiala, line 33: Only after_create? Then where do we store information about each time commission changes?

brianlong:
Looking at the after_create callback on the ValidatorHistory model, the additional data processing will cause performance problems and slow down the data gathering pipeline.

A different idea: After the end of the data gathering pipeline, create a Sidekiq job with the batch UUID. The background worker will then loop through each of the ValidatorHistory records in the batch and look for commission changes. When we detect a change, we can create the CommissionHistory record with the epoch info, etc. Basically, move from real-time to background processing so we can keep the data gathering pipeline as fast as possible.

brianlong:
FYI, I might have time to work on this task too. Please push your WIP at the end of each day so I know where you left off.

kbiala:
I'd use after_save callback in ValidatorScoreV1 record with condition `if commission_changed?` (rails built-in method). That would take much less time - no additional queries to find previous history, no comparison, because we already know if commission changed or not.

brianlong:
> I'd use after_save callback in ValidatorScoreV1 record with condition `if commission_changed?` (rails built-in method). That would take much less time - no additional queries to find previous history, no comparison of histories records, because we already know if commission changed or not.
> https://api.rubyonrails.org/v5.1.6/classes/ActiveRecord/AttributeMethods/Dirty.html#method-i-attribute_changed-3F

That could work and will not slow down the data gathering pipeline. Will we have the epoch number & completed % complete in scope to save with the commission change?

maciejkocylapc:
just added the changes advised by @kbiala and additional info about epochs. Please review

app/models/validator_score_v1.rb
```diff
@@ -78,6 +79,12 @@ class ValidatorScoreV1 < ApplicationRecord
     where(data_center_key: data_center_keys)
   end
 
+  def create_commission_history
+    if commission != commission_before_last_save
```
kbiala, line 83: This could also be shorter, as condition on callback, like this: `after_save :create_commission_history, if saved_change_to_commission?`. But this will also work.

app/services/create_commission_history_service.rb
```diff
@@ -0,0 +1,47 @@
...
+      commission_after: @score.commission,
+      batch_uuid: last_batch.uuid,
+      epoch: recent_epoch.epoch,
+      epoch_completion: recent_epoch_completion
+    )
+  end
+
+  def last_batch
+    @last_batch ||= Batch.where(network: @score.network)
+         .order(created_at: :desc)
```
kbiala, line 33: Wrong indentation

test/models/validator_score_v1_test.rb
```diff
@@ -74,4 +74,22 @@ class ValidatorScoreV1Test < ActiveSupport::TestCase
...
+    validator = create(:validator, network: 'testnet')
+    score = create(
+      :validator_score_v1,
+      validator: validator,
+      commission: 10,
+      network: 'testnet'
+    )
+    create(:epoch_history, network: 'testnet', batch_uuid: @batch.uuid)
+
+    refute CommissionHistory.exists?
```
kbiala, line 89: You can use assert_difference instead:
```
assert_difference('CommissionHistory.count') do
  score.update(commission: 20)
end
```

kbiala (approved):

PR #149, opened 2021-07-09
178811031 old commission changes

maciejkocylapc:
#### What's this PR do?
script to add historical commission changes

rails r script/back_fill_commission_history.rb

https://www.pivotaltracker.com/story/show/178811031

#### Things to watch out for


kbiala:
@maciejkocylapc We'll need to gather network as well. Please merge last updates from master (migration with network is already there) and update your script.

app/controllers/public_controller.rb
```diff
@@ -103,6 +103,18 @@ def contact_us
     @title = t('public.contact_us.title')
   end
 
+  def commission_histories
+    if params[:validator_id]
```
radek-polcode, line 107:
```suggestion
   commission_histories = if params[:validator_id]
```
and then you can remove assignments to `commission_histories` inside if-else.

app/models/batch.rb
```diff
@@ -26,6 +26,7 @@
 #
 # Indexes
 #
+#  index_batches_on_network                 (network)
```
brianlong, line 29: You don't need this index! You already have it with the indexes below.

app/models/commission_history.rb
```diff
@@ -8,13 +8,16 @@
...
 #  epoch_completion  :float(24)
+#  network           :string(191)
 #  created_at        :datetime         not null
 #  updated_at        :datetime         not null
 #  validator_id      :bigint           not null
 #
 # Indexes
 #
-#  index_commission_histories_on_validator_id  (validator_id)
+#  index_commission_histories_on_network                   (network)
```
brianlong, line 18: You don't need this index! You already have it with the indexes below.

app/models/commission_history.rb
```diff
@@ -8,13 +8,16 @@
...
+#  index_commission_histories_on_network_and_validator_id  (network,validator_id)
+#  index_commission_histories_on_validator_id              (validator_id)
```
brianlong, line 20: The three indexes above don't make sense when taken together. You'll never need to run a query on network + validator_id. Normally, you will have an index on [network, account] and then another index on validator_id.

app/views/validators/show.html.erb
```diff
@@ -123,7 +123,10 @@
               @validator.commission.to_i == 100
           %>
           <td class="<%= comm_class %>">
-            <%= @validator.commission rescue 'N/A' %> &percnt;
+            <%= @validator.commission rescue 'N/A' %>&percnt;
+            <%= link_to '(See commission changes)',
+                        commission_histories_path(network: params[:network], validator_id: @validator.id),
+                        class: 'small' %>
```
brianlong, line 129: Please make this link visible only when there are commission changes to show. It is bad UX for the user to click a link and then have nothing to show them.

app/controllers/public_controller.rb
kbiala: I don't think so.. Still need to assign @validator only if if params[:validator_id] is present

kbiala (approved):

PR #159, opened 2021-07-16
178911569 single vote account in charts

maciejkocylapc:
#### What's this PR do?
use last vote account in validator detail charts

see if charts are displayed as they should

https://www.pivotaltracker.com/story/show/178911569

#### Things to watch out for


brianlong:
@kbiala @maciejkocylapc Are you sure that simply using `Validator#vote_account_last` in the right place doesn't solve the problem? Where in our code are we grabbing both vote accounts when we only want the most recent?

PR #175, opened 2021-07-28
178551916 cluster stats frontend stage 2

kbiala:
#### What's this PR do?
- Cluster stats view, step 2

- deply to staging, go to /cluster-stats/mainnet and /cluster-stats/testnet

- https://www.pivotaltracker.com/n/projects/2177859/stories/178551916

#### Things to watch out for

### DIFF
--- a/app/views/cluster_stats/index.html.erb
+++ b/app/views/cluster_stats/index.html.erb
@@ -38,14 +38,14 @@
           </tr>
           </thead>
           <tbody>
-          <% @stats[:top_staked_validators].each.with_index do |stake_record, index| %>
+          <% @stats[:top_staked_validators].each.with_index do |record, index| %>
             <tr>
               <td><%= index + 1 %>.</td>
               <td>
-                <% validator = Validator.find(stake_record.last) %>
+                <% validator = Validator.find(record.last) %>
                 <%= link_to validator_display_name(validator), validator_path(network: params[:network], account: validator.account) %>
               </td>
-              <td><%= number_with_delimiter( lamports_to_sol(stake_record.first.to_i).round(0), delimiter: ',')  %></td>
+              <td><%= number_with_delimiter( lamports_to_sol(record.first.to_i).round(0), delimiter: ',')  %></td>
               <td class="text-right">
                 <%= link_to validator_path(network: params[:network], account: validator.account) do %>
                   <i class="fas fa-search text-purple"></i>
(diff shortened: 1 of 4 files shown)


app/views/cluster_stats/index.html.erb
```diff
@@ -88,12 +102,19 @@
...
           <tbody>
-          <% @stats[:top_root_distance_validators].first(50).each.with_index do |v, index| %>
+          <% @stats[:top_root_distance_validators].each.with_index do |record, index| %>
             <tr>
               <td><%= index + 1 %>.</td>
-              <td>Validator name</td>
-              <td><%= v.round(3) %></td>
-              <td class="text-right"><i class="fas fa-search text-purple"></i></td>
+              <td>
+                <% validator = Validator.find(record.last) %>
```
radek-polcode, line 109: Please get rid of this query from view

app/views/cluster_stats/index.html.erb
```diff
@@ -113,12 +134,19 @@
...
           <tbody>
-          <% @stats[:top_vote_distance_validators].first(50).each.with_index do |v, index| %>
+          <% @stats[:top_vote_distance_validators].each.with_index do |record, index| %>
             <tr>
               <td><%= index + 1 %>.</td>
-              <td>Validator name</td>
-              <td><%= v.round(3) %></td>
-              <td class="text-right"><i class="fas fa-search text-purple"></i></td>
+              <td>
+                <% validator = Validator.find(record.last) %>
```
radek-polcode, line 141: Please get rid of this query from view

app/views/cluster_stats/index.html.erb
```diff
@@ -138,12 +166,20 @@
...
           <tbody>
-          <% @stats[:top_skipped_slot_validators].first(50).each.with_index do |v, index| %>
+          <% @stats[:top_skipped_slot_validators].each.with_index do |record, index| %>
             <tr>
               <td><%= index + 1 %></td>
-              <td>Validator name</td>
-              <td><%= v.round(3) %></td>
-              <td class="text-right"><i class="fas fa-search text-purple"></i></td>
+              <td>
+                <% validator = Validator.find(record.last) %>
```
radek-polcode, line 173: Please get rid of this query from view

app/views/cluster_stats/index.html.erb
```diff
@@ -138,12 +166,20 @@
...
+                <%= link_to validator_display_name(validator), validator_path(network: params[:network], account: validator.account) %>
+              </td>
+              <td><%= record.first.to_f.round(3) %></td>
+              <td class="text-right">
+                <% validator = Validator.find(record.last) %>
```
radek-polcode, line 178: Please get rid of this query from view

radek-polcode (changes requested):
I think we are going in wrong direction with our query objects.

One of the main rule is that all methods should be chainable, so `ActiveRecord::Relation` should be returned, not array or hash.
Second, there should be no logic, all methods where we count sth should be moved to separate logic/service file.

Good explanation here: 
1. https://medium.flatstack.com/query-object-in-ruby-on-rails-56ea434365f0
2. https://www.ironin.it/blog/design-patterns-in-large-rails-applications-query-objects.html

brianlong:
> I think we are going in wrong direction with our query objects.
> 
> One of the main rule is that all methods should be chainable, so `ActiveRecord::Relation` should be returned, not array or hash.
> Second, there should be no logic, all methods where we count sth should be moved to separate logic/service file.
> 
> Good explanation here:
> 
> 1. https://medium.flatstack.com/query-object-in-ruby-on-rails-56ea434365f0
> 2. https://www.ironin.it/blog/design-patterns-in-large-rails-applications-query-objects.html

+1 to this. It seems like most query logic should be contained inside the ActiveRecord model. I don't like jumping around a bunch of different files to understand how something works. The query objects might be OK where you are querying multiple tables and it doesn't make sense to put the logic inside of the related ActiveRecord models.

Also, remember that one of the goals for this project is to explore the functional side of Ruby.

kbiala:
@radek-polcode I completely agree. As we discussed on slack, refactoring this code would take some time, and is way out of the scope of this story, so I created a follow-up story here: https://www.pivotaltracker.com/n/projects/2177859/stories/179117102, so that we can push this story. Please take another look at my changes.

radek-polcode (approved):

PR #215, opened 2021-09-22
179475514 flag for absent validators in api

maciejkocylapc:
#### What's this PR do?
add a bolean parameter to validators list api endpoint to exclude validators that are no longer returned by solana cli

- run tests
- check if api endpoint works as expected (/api/v1/validators/testnet?show_absent=false)
- (you can remove validator_history for a validator, after that it should not be seen if show_absent attribute is set to false)

https://www.pivotaltracker.com/story/show/179475514

#### Things to watch out for


radek-polcode (changes requested):
Please fix performance

PR #224, opened 2021-10-01
179752479_add_delinquent_stats_to_data_center_views

radek-polcode:
#### What's this PR do?
- add delinquent stats to data center views

- Go to data centers, see the delinquent column

- https://www.pivotaltracker.com/story/show/179752479

#### Things to watch out for

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -97,6 +97,7 @@ en:
     asn_full_name: "Autonomous System Number"
     average_score: "Average score for validators from ASN."
     commission: "Commission is the percent of network rewards earned by a validator that are deposited into the validator's vote account."
+    delinquent: "Delinquent definition."
     isp_full_name: "Internet Service Provider"
     population_asn: "Number of validators for ASN."
     population_isp: "Number of validators for ISP."
--- a/app/helpers/data_centers_helper.rb
+++ b/app/helpers/data_centers_helper.rb
@@ -14,4 +14,8 @@ module DataCentersHelper
       render 'data_center_list'
     end
   end
+
+  def count_percentage(dividend, divisor)
+    (dividend.to_f / divisor.to_f * 100).round(2)
+  end
 end
--- a/app/services/sorted_data_centers.rb
+++ b/app/services/sorted_data_centers.rb
@@ -7,6 +7,7 @@ class SortedDataCenters
     {
       total_population: @total_population,
       total_stake: @total_stake,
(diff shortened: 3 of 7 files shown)


maciejkocylapc (changes requested):

maciejkocylapc (approved):

config/locales/en.yml
```diff
@@ -97,6 +97,7 @@ en:
     asn_full_name: "Autonomous System Number"
     average_score: "Average score for validators from ASN."
     commission: "Commission is the percent of network rewards earned by a validator that are deposited into the validator's vote account."
+    delinquent: "Delinquent validator is a validator that does not give rewards for the epoch for some period of time."
```
kbiala, line 100:
```suggestion
    delinquent: "Delinquent validator is a validator that has 100% of commission in epoch."
```

kbiala (changes requested):
Since you added a new column, table became wider, and content no longer fits on smaller screens. You need to change `table-responsive-md` class to `table-responsive-md` in data center index view.

config/locales/en.yml
radek-polcode: Do you have some source of this defintion? I don't think it's correct.

kbiala: There's no external source - this is the term that we (validators) introduced.

kbiala: Actually, those are private validator's definition (both mine and yours). You need to check conditions in code for delinquent validato. There two are not correct.

radek-polcode: We get delinquent field value from CLI request, I can't see any modification made by our code - https://github.com/brianlong/www.validators.app/blob/master/app/logic/solana_logic.rb#L151

Why it's private validator's definitions?

kbiala: `does not give rewards` - this is about private validators, they keep reward. From what I can see, there are many cases in which Solana can mark validator as delinquent, so I don't think it's a good idea to explain it here. Also, I'd use something more informative for the context in which this definition is used, so that the users know what this column represents in this particular case. So instead, something like:
```Numer of delinquent nodes in Data Center```

radek-polcode: Where did you find this information about marking validator as delinquent by solana? I can't find it anywhere.

kbiala: I couldn't find any direct clear explanation on Solana website either, which is a pity. I switched to Discord some time ago, there's a lot more information there.

PR #237, opened 2021-10-15
179919622 appsignal validator show page

maciejkocylapc:
#### What's this PR do?
rescue the RecordNotFound errors

try to visit validator's page that does not exist

- https://www.pivotaltracker.com/story/show/179919622
- https://www.pivotaltracker.com/story/show/179919659

#### Things to watch out for

### DIFF
--- a/app/controllers/validators_controller.rb
+++ b/app/controllers/validators_controller.rb
@@ -114,9 +114,7 @@ class ValidatorsController < ApplicationController
     @validator = Validator.where(
       network: params[:network],
       account: params[:account]
-    ).first or not_found
-  rescue ActiveRecord::RecordNotFound
-    render_404
+    ).first or render_404
   end
 
   def validate_order


app/controllers/public_controller.rb
```diff
@@ -104,7 +104,11 @@ def contact_us
 
   def commission_histories
     if params[:validator_id]
-      @validator = Validator.find(params[:validator_id])
+      begin
+        @validator = Validator.find(params[:validator_id])
+      rescue ActiveRecord::RecordNotFound
```
psuchta, line 109: We have to rescue the error ? I think that when the  RecordNotFound occurs Rails goes to render_404 by default.

app/controllers/validators_controller.rb
```diff
@@ -115,6 +115,8 @@ def set_validator
       network: params[:network],
       account: params[:account]
     ).first or not_found
+  rescue ActiveRecord::RecordNotFound
```
psuchta, line 118: Same as above

app/controllers/public_controller.rb
maciejkocylapc: I'm rescuing the error so that appsignal does not report unnecessary information

app/controllers/validators_controller.rb
psuchta: One additional question here - Can we do that ? 
```suggestion
 @validator = Validator.where(
      network: params[:network],
      account: params[:account]
    ).first or render_404
```

PR #258, opened 2021-11-05
180178401_no_commission_changes_visible

radek-polcode:
#### What's this PR do?
- ADd correct checking for commission historiesbefore displaying link

- Deploy to staging and check the links from the description in PT story (or try to do it on localhost)

- https://www.pivotaltracker.com/story/show/180178401

#### Things to watch out for

### DIFF
--- a/app/controllers/validators_controller.rb
+++ b/app/controllers/validators_controller.rb
@@ -69,7 +69,10 @@ class ValidatorsController < ApplicationController
     # Grab the distances to show on the chart
     @vote_blocks = @val_histories.map(&:vote_distance).compact
 
-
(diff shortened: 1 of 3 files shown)


app/controllers/validators_controller.rb
```diff
@@ -69,7 +69,10 @@ def show
     # Grab the distances to show on the chart
     @vote_blocks = @val_histories.map(&:vote_distance).compact
 
-
+    @commission_histories = CommissionHistoryQuery.new(
```
kbiala, line 72: I think .count would be enough in this case. We're not displaying histories here, just checking if they exist.

kbiala (approved):

PR #267, opened 2021-11-17
180143125-cache-cluster-stats

maciejkocylapc:
#### What's this PR do?
- create report with cluster stats and use it for the views

- run gather rpc daemon and turn on sidekiq
- check if report was created with the name 'report_cluster_stats'
- check if cluster stats view works as expected

https://www.pivotaltracker.com/story/show/180143125

#### Things to watch out for


app/logic/report_logic.rb
```diff
@@ -243,4 +243,46 @@ def report_software_versions
...
+
+  def report_cluster_stats
+    lambda do |p|
+      batch = Batch.find_by(uuid: p.payload[:batch_uuid])
+      network = p.payload[:network]
+
+      vah_stats = Stats::VoteAccountHistory.new(network, batch.uuid)
+      vbh_stats = Stats::ValidatorBlockHistory.new(network, batch.uuid)
+      vs_stats  = Stats::ValidatorScore.new(network, batch.uuid)
+      software_report = Report.where(
```
radek-polcode, line 255: empty line above

radek-polcode (changes requested):
Report for testnet didn't generate. Stats for software versions are empty or zeros.

radek-polcode (approved):

kbiala (commented):
I can't see removing old versions anywhere. We should not need more than 2-3 versions for each network

app/views/validators/_software_versions.html.erb
```diff
@@ -18,14 +18,14 @@
               <% sort_software_versions(software_versions).each do |group| %>
                 <% if group.first[1].is_a?(Hash) %>
                   <tr>
-                    <td class="<%= group.first[0] == current_software_version(batch, params[:network]) ? 'h4 text-success' : '' %>">
+                    <td class="<%= group.first[0].to_s == current_software_version(batch, params[:network]).to_s ? 'h4 text-success' : '' %>">
```
kbiala, line 21: There's no batch variable anymore (see changes you made to controller), so this version is not displayed correctly. You should use batch number saved in report. Check also other references to batch in cluster stats views.

kbiala (changes requested):
I'm still worried that clean up that we currently have (older than a month) is not enough in his case - we generate these reports every few minutes (x2 networks), so it would be a lot of new records in db.

app/views/validators/_software_versions.html.erb
maciejkocylapc: here's the batch assingment:
https://github.com/brianlong/www.validators.app/pull/267/files#diff-a777acc0360af07189fc742f4cb67caba42a8db3f17fda4bbdbe79fa37a4c510R9

app/workers/report_cluster_stats_worker.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+# ReportClusterStatsWorker.perform_async(
+#   network: 'testnet',
+#   batch_uuid: '65c9083f-8f53-4873-a9f8-8f782e276d30'
+# )
```
kbiala, line 6: Do we need this comment?

app/workers/report_cluster_stats_worker.rb
```diff
@@ -0,0 +1,21 @@
...
+class ReportClusterStatsWorker
+  include Sidekiq::Worker
+  include ReportLogic
+
+  def perform(args = {})
+    # byebug
```
kbiala, line 12: You can remove this line

kbiala (approved):

PR #288, opened 2021-12-09
180570785_change_price_format

radek-polcode:
#### What's this PR do?
- Uses comma as a delimiterfor thousands + display only two decimal places
- 
- Go to `localhost:3000/stake_accounts` and check how sols in first columns are displayed (they should be delimited by thousands and have 2 decimal places)

- https://www.pivotaltracker.com/story/show/180570785

#### Things to watch out for

### DIFF
--- a/app/models/stake_account.rb
+++ b/app/models/stake_account.rb
@@ -37,11 +37,4 @@ class StakeAccount < ApplicationRecord
   scope :filter_by_account, ->(account) { where('stake_pubkey LIKE ?', "#{account}%") }
   scope :filter_by_staker, ->(staker) { where('staker LIKE ?', "#{staker}%") }
   scope :filter_by_withdrawer, ->(withdrawer) { where('withdrawer LIKE ?', "#{withdrawer}%") }
-
-  def formatted_delegated_stake
-    return "0.00" unless delegated_stake.is_a?(Numeric)
-
-    stake_in_sol = (delegated_stake / 1_000_000_000.to_f)
-    ActiveSupport::NumberHelper.number_to_currency(stake_in_sol, precision: 2, unit: "")
-  end
 end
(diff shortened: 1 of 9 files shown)


maciejkocylapc (changes requested):
There is no delimiter for thousands

app/models/stake_account.rb
```diff
@@ -37,4 +37,11 @@ class StakeAccount < ApplicationRecord
   scope :filter_by_account, ->(account) { where('stake_pubkey LIKE ?', "#{account}%") }
   scope :filter_by_staker, ->(staker) { where('staker LIKE ?', "#{staker}%") }
   scope :filter_by_withdrawer, ->(withdrawer) { where('withdrawer LIKE ?', "#{withdrawer}%") }
+
+  def formatted_delegated_stake
+    return "0.00" unless delegated_stake.is_a?(Numeric)
+
+    stake_in_sol = (delegated_stake / 1_000_000_000.to_f)
```
kbiala, line 44: Generally ok, but I'd make it a helper as well, it may be useful in case we want to transform lamports to sols in some other place.

test/helpers/application_helper_test.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+class ApplicationHelperTest < ActiveSupport::TestCase
+  include ApplicationHelper
+
+  test "#lamports_as_formatted_sol returns formatted sol with delimiters" do
+    assert_equal "12,341.23", lamports_as_formatted_sol(12341232212414)
```
maciejkocylapc, line 9: You could also test for other formats like nil or string

maciejkocylapc (approved):

PR #294, opened 2021-12-20
180640347 mark removed validators

maciejkocylapc:
#### What's this PR do?
- mark validators that are no longer in solana as destroyed

- run tests
- try to run `ValidatorCheckActiveService.new.update_validator_activity` in rails console after running daemons for a while, although it's not certain that there will be any marked
- already tested on stage and there are some validators marked 

[- URL to the PivotalTracker ticket](https://www.pivotaltracker.com/story/show/180640347)

#### Things to watch out for

### DIFF
--- a/app/services/validator_check_active_service.rb
+++ b/app/services/validator_check_active_service.rb
@@ -30,9 +30,9 @@ class ValidatorCheckActiveService
   private
 
   def should_be_destroyed?(validator)
-    if validator.validator_histories.where("created_at > ?", @delinquent_time)
-      if validator.validator_histories.where('created_at < ?', @delinquent_time).exists?
-        true
+    if !validator.validator_histories.where("created_at < ?", @delinquent_time).exists?
+      if validator.validator_histories.where('created_at > ?', @delinquent_time).exists?
+        return true
       end
     end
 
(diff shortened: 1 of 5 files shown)


app/services/validator_check_active_service.rb
```diff
@@ -27,6 +29,16 @@ def update_validator_activity
 
   private
 
+  def should_be_destroyed?(validator)
+    if !validator.validator_histories.where("created_at > ?", @delinquent_time.ago).exists?
+      if validator.validator_histories.where('created_at < ?', @delinquent_time.ago).exists?
```
kbiala, line 34: Why nested instead of &&? Also, this type of negation in this case is not very clear, it's easy to miss, I'd use `empty?` instead.

app/services/validator_check_active_service.rb
```diff
@@ -10,7 +10,9 @@ def initialize
 
   def update_validator_activity
     Validator.all.each do |validator|
```
radek-polcode, line 12: `includes(:vote_accounts)` to decrease number of db queries

app/services/validator_check_active_service.rb
```diff
@@ -27,6 +29,16 @@ def update_validator_activity
...
+        return true
+      end
+    end
+
+    false
+  end
+
   # number of previous by network
   def previous_epoch(network)
     current_epoch = EpochWallClock.where(network: network).order(created_at: :desc).last
```
radek-polcode, line 44: If it's not intentionally this also could be cached to not generate db queries every time.

radek-polcode (approved):

kbiala (approved):

PR #343, opened 2022-02-07
181163219_show_geo_data_in_api

radek-polcode:
#### What's this PR do?
- Adds latitude and longitude to api response + update the docs

- Use postmant or curl, request validators list or single validator (make sure it has ip record connected) and check if latitude and longitude are present

- https://www.pivotaltracker.com/story/show/181163219

#### Things to watch out for


test/controllers/api/v1/api_controller_test.rb
```diff
@@ -149,7 +149,10 @@ def setup
     assert_nil validator_with_all_data["ping_time"]
 
     # IP
```
psuchta, line 151: Please add a test for checking if `latitude` and `longitude` return values when they are not nil

psuchta (changes requested):
Please add a test for checking if `latitude` and `longitude` return values when they are not nil

psuchta (approved):

PR #359, opened 2022-02-16
181242369 add ping thing in batches

maciejkocylapc:
#### What's this PR do?
- create new endpoint for adding multiple transactions to ping_thing

- try posting multiple records to `/api/v1/ping-thing-batch/testnet`

- [URL to the PivotalTracker ticket](https://www.pivotaltracker.com/story/show/181242369)

#### Things to watch out for


app/controllers/api/v1/ping_things_controller.rb
```diff
@@ -18,11 +20,37 @@ def create
         end
       end
 
+      def create_multiple
+        api_token = request.headers["Token"]
+
+        begin
+          raise InvalidRecordCount, "Too many records" \
```
radek-polcode, line 27: It would be nice to add max number of records to this message.

config/routes.rb
```diff
@@ -140,6 +140,7 @@
 
       # POST /api/vi/ping-thing/
       post 'ping-thing/:network', to: 'ping_things#create', as: 'ping_thing'
+      post 'ping-thing-batch/:network', to: 'ping_things#create_multiple', as: 'ping_thing_batch'
```
radek-polcode, line 143:
```suggestion
      post 'ping-things/batch/:network', to: 'ping_things#create_multiple', as: 'ping_thing_batch'
```
I am still concerned about using singular paths for resources instead of plurals, it does not follow REST API conventions. Singular form would be ok if we have only one ping thing.

https://stackoverflow.com/a/21809963/3868595
https://stackoverflow.com/questions/6845772/should-i-use-singular-or-plural-name-convention-for-rest-resources#comment52501175_6846057
https://restfulapi.net/resource-naming/

app/controllers/api/v1/ping_things_controller.rb
```diff
@@ -18,11 +20,37 @@ def create
         end
       end
 
+      def create_multiple
```
radek-polcode, line 23:
```suggestion
      def create_batch
```
to be consistent with the route?

app/views/public/api_documentation.html.erb
```diff
@@ -364,9 +365,29 @@
...
+</div>
+
+<hr />
+
+<h4 id="ping-thing-batch">Ping Thing Batch</h4>
+<div class="row">
+  <div class="col-sm-12">
+    <p>The ping thing batch endpoint allows you to push multiple transaction informations at once.</p>
+    <p>Request:</p>
+    <pre>curl -X POST -H "Token: secret-api-token" -H "Content-Type: application/json" -d '{"transactions": ["time": "NN", "signature": "signature", "transaction_type": "transfer"}]'
```
radek-polcode: `reported_at` is missing, please add it also above. I think we should stick to ISO 8601 format as it is a general consensus for sending datetimes in json https://stackoverflow.com/a/15952652/3868595

Also I think you missed  curly brace  before "time" and at the end of the object inside array. There is also wrong order at the end. It would be nice to provide at least two objects in the example array since it's a batch.

test/controllers/api/v1/ping_thing_controller_test.rb
```diff
@@ -43,4 +43,45 @@ def setup
...
+  test "create_multiple request with token should return 200 and save the data" do
+    params_batch = {
+      transactions: [
+        @params_sample, @params_sample
+      ]
+    }
+
+    post api_v1_ping_thing_batch_path(
+      network: "testnet"
+    ), headers: { "Token" => @user.api_token }, params: params_batch
```
radek-polcode, line 64:
```suggestion
    ), headers: { "Token" => @user.api_token }, params: params_batch.to_json
```
I think it would be more appropriate to send params in json format

test/controllers/api/v1/ping_thing_controller_test.rb
```diff
@@ -43,4 +43,45 @@ def setup
...
+    assert_response 201
+    expected_response = { "status" => "created" }
+
+    assert_equal expected_response, response_to_json(@response.body)
+    assert_equal params_batch[:transactions].last, JSON.parse(PingThingRaw.last.raw_data).symbolize_keys
+  end
+
+  test "create_multiple request with token should return 400 when there's too much data" do
+    transactions = []
+    51.times {transactions.push(@params_sample)}
```
radek-polcode, line 74: I think we agreed on higher limit

app/controllers/api/v1/ping_things_controller.rb
```diff
@@ -32,14 +34,41 @@ def create
         end
       end
 
+      def create_batch
+        max_records = 50
```
radek-polcode, line 38: I think we agreed on higher number, it might also be a constant

app/views/public/api_documentation.html.erb
```diff
@@ -359,6 +360,22 @@ curl -X POST -H "Token: secret-api-token" -H "Content-Type: application/json" -d
 
 <hr />
 
+<h4 id="ping-thing-batch">Ping Thing Post Batch</h4>
+
+<p>The ping thing batch endpoint allows you to push multiple transaction informations at once.</p>
+<p>Request:</p>
+<pre>curl -X POST -H "Token: secret-api-token" -H "Content-Type: application/json" -d '{"transactions": ["time": "NN", "signature": "signature", "transaction_type": "transfer"}]'
```
radek-polcode, line 367:
```suggestion
<pre>curl -X POST -H "Token: secret-api-token" -H "Content-Type: application/json" -d '{"transactions": [{ "time": "NN", "signature": "signature", "transaction_type": "transfer" }, { second_example } ]}'
```

`reported_at` is missing. Also it would be nice to add second example since it's a batch. We should stick to ISO 8601 format https://stackoverflow.com/a/15952652/3868595

radek-polcode (changes requested):

radek-polcode (approved):

kbiala:
@maciejkocylapc Please revolve conflicts (and update your code with current master if necessary)

PR #368, opened 2022-02-24
181341014 sort pings by reported at

kbiala:
#### What's this PR do?
- Uses reported_at instead of created_at

- Review in browser under localhost:3000/ping-thing or on staging
- Try fetching more than one ping by API and confirm that they're sorted by reported_at
eg. pull by ```curl -H "Token: YOUR-KEY" 'https://stage.validators.app/api/v1/ping-thing/mainnet?limit=5```
and compare results with: https://stage.validators.app/ping-thing

- [URL to the PivotalTracker ticket](https://www.pivotaltracker.com/n/projects/2177859/stories/181341014)

#### Things to watch out for

### DIFF
--- a/app/models/ping_thing.rb
+++ b/app/models/ping_thing.rb
@@ -22,7 +22,7 @@
 #
 #  index_ping_things_on_created_at_and_network_and_transaction_type  (created_at,network,transaction_type)
 #  index_ping_things_on_created_at_and_network_and_user_id           (created_at,network,user_id)
-#  index_ping_things_on_reported_at                                  (reported_at)
+#  index_ping_things_on_reported_at_and_network                      (reported_at,network)
 #  index_ping_things_on_user_id                                      (user_id)
 #
 # Foreign Keys
--- a/app/models/ping_thing.rb
+++ b/app/models/ping_thing.rb
@@ -50,7 +50,8 @@ class PingThing < ApplicationRecord
         :response_time,
         :signature,
         :success,
-        :transaction_type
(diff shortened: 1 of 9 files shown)


db/schema.rb
```diff
@@ -358,17 +358,17 @@
     t.string "network"
     t.datetime "created_at", precision: 6, null: false
     t.datetime "updated_at", precision: 6, null: false
-    t.float "manager_fee"
     t.string "ticker"
+    t.float "manager_fee"
```
maciejkocylapc, line 362: this should not be here

db/schema.rb
```diff
@@ -358,17 +358,17 @@
...
     t.float "average_validators_commission"
     t.float "average_delinquent"
     t.float "average_skipped_slots"
     t.float "average_uptime"
     t.integer "average_lifetime"
     t.float "average_score"
+    t.float "average_apy"
```
maciejkocylapc, line 369: this should not be here

app/models/ping_thing.rb
```diff
@@ -22,7 +22,7 @@
 #
 #  index_ping_things_on_created_at_and_network_and_transaction_type  (created_at,network,transaction_type)
 #  index_ping_things_on_created_at_and_network_and_user_id           (created_at,network,user_id)
```
maciejkocylapc, line 24: perhaps we should delete this index if it's not used anymore

maciejkocylapc (changes requested):

PR #377, opened 2022-03-07
181482482_fix_query_in_sorted_data_centers

radek-polcode:
#### What's this PR do?
- Add missing condition to query in sorted_data_centers service

- Switch between branch master and this one
- Check if the numbers are the same (or you may see some differences - that means that new change in query works, find differenece and compare)

- https://www.pivotaltracker.com/story/show/181482482

#### Things to watch out for


kbiala (changes requested):
Code loooks good, but the page loads for a very long time (on staging). Also query alone called from the console loads for a very long time. Please check the query, indexes etc.

radek-polcode:
After switching to data_centers this won't be needed.

PR #386, opened 2022-03-15
181570523_adjust_api_to_data_centers

radek-polcode:
#### What's this PR do?
- Adjust API to use data_centers instead of ips

- Check API responses (data and performance; best to do that locally is to switch between branch master and this one)
- Best QA is possible on staging since it has all the data needed (all data_centers)

- https://www.pivotaltracker.com/story/show/181570523

#### Things to watch out for


test/factories/validators.rb
```diff
@@ -18,5 +18,17 @@
...
       network { 'mainnet' }
     end
+
+    trait :with_validator_ip do
+      after(:create) do |validator|
+        create :validator_ip, :active, validator: validator
+      end
+    end
+
+    trait :with_validator_ip_active_and_data_center do
```
maciejkocylapc, line 28: I'd change that to with_data_center_through_validator_ip

maciejkocylapc (approved):

test/models/validator_test.rb
```diff
@@ -47,37 +77,38 @@ def setup
   end
 
   test '#private_validator? returns true if validator\'s commission is 100%' do
-    validator = create(:validator, network: 'mainnet')
-    assert create(:validator_score_v1, commission: 100, validator: validator).validator.private_validator?
-    refute create(:validator_score_v1, commission: 99, validator: validator).validator.private_validator?
+    @validator.update(network: 'mainnet')
```
maciejkocylapc, line 80: perhaps it's better to create separate validator as it can potentially impact other tests

test/models/validator_test.rb
```diff
@@ -47,37 +77,38 @@ def setup
...
+    refute @validator.lido?
   end
 
-  test 'validator attributes are correctly stored in db after using utf_8_encode method' do
-    special_chars_sentence = 'Staking-P◎◎l FREE Validati◎n'
+  test "validator attributes are correctly stored in db after using utf_8_encode method" do
+    special_chars_sentence = "Staking-P◎◎l FREE Validati◎n"
 
-    v = Validator.create(
+    @validator.update(
```
maciejkocylapc, line 100: same as above

test/queries/validator_query_test.rb
```diff
@@ -58,13 +88,19 @@ class ValidatorQueryTest < ActiveSupport::TestCase
...
     end
 
     result = ValidatorQuery.new.call(network: @mainnet_network, sort_order: "name")
 
     assert_equal 5, result.count
-    assert_equal (0..4).map{|n| "#{n}-validator"}, result.pluck(:name)
+    assert_equal (0..4).map{ |n| "#{n}-validator" }, result.pluck(:name)
   end
 
 end
```
maciejkocylapc, line 106: add empty line

maciejkocylapc (changes requested):
There is no tests for when validator has no data_center assigned to it. Does it appear in API response?

app/controllers/concerns/validators_controller_helper.rb
```diff
@@ -6,17 +6,19 @@ def create_json_result(validator, with_history = false)
 
     hash.merge!(validator.to_builder.attributes!)
 
-    score = validator.score
-    ip = score.ip_for_api if score
-    vote_account = validator.vote_accounts.last
+    validator_score = with_history ? validator.validator_score_v1 : validator.validator_score_v1_for_api
+
+    data_center = validator.validator_ip_active_for_api&.data_center_host_for_api&.data_center_for_api
```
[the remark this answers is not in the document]
radek-polcode: This works (does not trigger n+1 queries) only if chained models are included in sql query, so I think we could leave it for now only here since query is only used here. You can use `validator.data_center_for_api` when need data_center for single validator.

test/models/validator_test.rb
radek-polcode: `setup` method is called before every test, so the initial validator is created and the changes from other tests are rollbacked. The easiest way to see it is to put ie. `puts @validator.network` at the bottom of the setup method and run tests.

maciejkocylapc (approved):

PR #394, opened 2022-03-25
181656771 validator block history new path

maciejkocylapc:
#### What's this PR do?
- change underscores to dashes in validator_block_history_path

- run tests
- check if `/api/v1/validator-block-history` endpoint works correctly
- check if the old path works as well

- [URL to the PivotalTracker ticket](https://www.pivotaltracker.com/story/show/181656771)

#### Things to watch out for

### DIFF
--- a/test/controllers/api/v1/validator_block_histories_controller_test.rb
+++ b/test/controllers/api/v1/validator_block_histories_controller_test.rb
@@ -33,6 +33,17 @@ class ValidatorsControllerTest < ActionDispatch::IntegrationTest
     assert_equal "0.25", json_response["skipped_slot_percent"]
   end
 
+  test "should redirect correctly if old url is given" do
+    validator = create(:validator, :with_score, account: "TestAccount")
+    create(:vote_account, validator: validator)
+    create(:validator_block_history, validator: validator)
+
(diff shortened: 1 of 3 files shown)


test/controllers/api/v1/validator_block_histories_controller_test.rb
```diff
@@ -33,6 +33,17 @@ class ValidatorsControllerTest < ActionDispatch::IntegrationTest
...
+    create(:vote_account, validator: validator)
+    create(:validator_block_history, validator: validator)
+
+    get "/api/v1/validator_block_history/testnet/#{validator.account}"
+
+    assert_response 301
+    assert_redirected_to api_v1_validator_block_history_path(account: validator.account, network: "testnet")
+  end
+
   test "GET api_v1_validator_block_history with token returns only validators from chosen network and validator" do
```
kbiala, line 47: I would update this test description to reflect the new URL format (and in other tests for this URL).

kbiala: Please ignore, looks like we're using helper name everywhere, so should be already good.

kbiala (approved):

PR #434, opened 2022-05-09
182036856 vote accounts is active

maciejkocylapc:
#### What's this PR do?
- add is_active flag for vote_accounts and set it in VoteAccountDetailsService

- run tests
- run migrations
- run gather_vote_account_details_daemon.rb
- check if there are accounts with is_active flag set to false

- [URL to the PivotalTracker ticket](https://www.pivotaltracker.com/story/show/182036856)
- https://www.pivotaltracker.com/story/show/182072088

#### Things to watch out for


radek-polcode (changes requested):

app/services/gatherers/vote_account_details_service.rb
```diff
@@ -9,25 +9,39 @@ class VoteAccountDetailsService
...
+                 .each do |vacc|
+        @range_start = vacc.id
+
         vote_account_details = get_vote_account_details(vacc.account)
-        unless vote_account_details.blank?
-          vacc.update(
+        if vote_account_details.blank?
+          vacc.set_inactive
+        else
+          if vacc.update(
```
radek-polcode, line 26: You should check here if the values from solana are different than what we already have nad based on that update or not the record

app/models/vote_account.rb
```diff
@@ -37,6 +38,12 @@ class VoteAccount < ApplicationRecord
   before_save :set_network
 
   scope :for_api, -> { select(FIELDS_FOR_API).order(updated_at: :asc) }
+  scope :active, -> { where(is_active: true) }
+
+  def set_inactive
+    self.is_active = false
+    self.save(touch: false)
```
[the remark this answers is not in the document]
maciejkocylapc: we never want to change timestamps if is_active is changed to false, but I've changed the method name to touch_inactive instead.

app/services/gatherers/vote_account_details_service.rb
```diff
@@ -6,6 +6,11 @@ module Gatherers
   class VoteAccountDetailsService
     include SolanaLogic
 
+    SCORE_OPTIONS = {
```
radek-polcode, line 9: I'd put that to the model and freeze that.

radek-polcode (changes requested):

app/services/gatherers/vote_account_details_service.rb
```diff
@@ -36,6 +45,11 @@ def call
 
     private
 
+    def should_omit_update?(vacc, vote_account_details)
+      vacc.validator_identity == vote_account_details["validatorIdentity"] && \
+      vacc.authorized_withdrawer == vote_account_details["authorizedWithdrawer"]
```
radek-polcode, line 50: Does this make sense to check `is_active` as well?

test/services/gatherers/vote_account_details_service_test.rb
```diff
@@ -0,0 +1,62 @@
...
+      validatorIdentity: @validator_identity,
+      authorizedWithdrawer: @authorized_withdrawer
+    }.to_json
+
+    @val = create(:validator, network: @network)
+    @score = create(:validator_score_v1, network: @network, validator: @val)
+  end
+
+  teardown do
+    VoteAccount.delete_all
```
radek-polcode, line 22: Is that needed?

test/services/gatherers/vote_account_details_service_test.rb
```diff
@@ -0,0 +1,62 @@
...
+      assert_equal 0, @score.reload.authorized_withdrawer_score
+    end
+  end
+
+  test "VoteAccountDetailsService sets correct score" do
+    va1 = create(
+      :vote_account,
+      validator: @val,
+      network: @network,
+      account: "acc1",
```
radek-polcode, line 46: I think this could be initialized in setup so you can use it in all tests, and two attributes below you can update here in test.

radek-polcode (changes requested):

radek-polcode (changes requested):
I think that we also need to test case when solana response is empty, if should_omit is true and if record update fails.

test/services/gatherers/vote_account_details_service_test.rb
```diff
@@ -0,0 +1,96 @@
...
+      Gatherers::VoteAccountDetailsService.new(network: @network, config_urls: [@testnet_url]).call
+
+      assert_equal @va1.reload.validator_identity, @validator_identity
+      assert_equal @va1.reload.authorized_withdrawer, @authorized_withdrawer
+      assert_equal 0, @score.reload.authorized_withdrawer_score
+    end
+  end
+
+  test "call sets negative score correctly" do
+    @json_response = {
```
radek-polcode, line 39: Does this need to be instance variable? Similar in other tests.

radek-polcode (changes requested):

kbiala (approved):

PR #453, opened 2022-05-31
182271924 stakeblaze pool

maciejkocylapc:
#### What's this PR do?
- add new stake pool

- test on staging
- check if there are logos and links for the new pool
- check if gathered data makes sense
- check if apy is correct

- [URL to the PivotalTracker ticket](https://www.pivotaltracker.com/story/show/182271924)

#### Things to watch out for


radek-polcode (approved):
Maybe it should display always three logos shuffled on the main page instead of one when we have number of stake pools that is not multiplication of 3?
Also maybe there is a way to make stake blaze iamge a bit bigger? it differs from the rest

kbiala:
Running on staging. Minor things to fix:
image
image

the name of the pools is BlazeStake, not Blazestake

PR #463, opened 2022-06-06
182343411-remove-public-index

maciejkocylapc:
#### What's this PR do?
- remove public#index since it duplicates validators#index

- check if links to home page are working correctly
- check if home page is working as it should
- run tests

- [URL to the PivotalTracker ticket](https://www.pivotaltracker.com/story/show/182343411)

#### Things to watch out for

### DIFF
--- a/app/controllers/validators_controller.rb
+++ b/app/controllers/validators_controller.rb
@@ -9,27 +9,31 @@ class ValidatorsController < ApplicationController
   def index
     @per = 25
 
-    if params[:watchlist]
+    if index_params[:watchlist]
       user_id = current_user&.id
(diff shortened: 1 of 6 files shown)


app/controllers/validators_controller.rb
```diff
@@ -126,4 +130,8 @@ def validate_order
 
     params[:order]
   end
+
+  def index_params
+    params.permit(:watchlist, :network, :q, :page)
```
kbiala, line 135: Looks like :order param is missing.

kbiala (changes requested):
Sorting validators is not working (they are always sorted by score).

kbiala (approved):

PR #492, opened 2022-06-29
182558082 1 hour stats with separate table

kbiala:
#### What's this PR do?
- Creates a new model for recent stats

- Run migrations
- Run sidekiq
- Run `rails r dev/create_test_ping_things.rb`
- Run `rails r script/one_time_scripts/back_fill_ping_thing_stats.rb`
- Check /ping-thing stat
- Create some more stats (eg. via console) and run `PingThingRecentStatsWorker.perform_async('testnet')`
- Check /ping-thing stat again and confirm that 5 and 60 min stats got updated

Staging:
- Test the same on staging, add new ping things via API
- Confirm that 5 and 60 min stats are updated once a minute (you need to reload the page to see the difference)

- [URL to the PivotalTracker ticket](https://www.pivotaltracker.com/n/projects/2177859/stories/182558082)

#### Things to watch out for

### DIFF
--- a/app/models/ping_thing.rb
+++ b/app/models/ping_thing.rb
@@ -40,7 +40,7 @@ class PingThing < ApplicationRecord
   validates :network, inclusion: { in: %w(mainnet testnet) }
   validates :signature, length: { in: 64..128 }
 
-  scope :for_date_range_and_network, -> (network, from, to) {
+  scope :for_reported_at_range_and_network, -> (network, from, to) {
     where(network: network, reported_at: (from..to))
   }
 
--- a/test/models/ping_thing_recent_stat_test.rb
+++ b/test/models/ping_thing_recent_stat_test.rb
@@ -5,10 +5,6 @@ require "test_helper"
 class PingThingRecentStatTest < ActiveSupport::TestCase
   include ActionCable::TestHelper
 
-  setup do
-
-  end
-
   test "doesn't create record with invalid interval" do
     stat = build(:ping_thing_recent_stat, interval: 12)
 
--- a/app/models/ping_thing_recent_stat.rb
+++ b/app/models/ping_thing_recent_stat.rb
@@ -35,7 +35,7 @@ class PingThingRecentStat < ApplicationRecord
   validates :interval, inclusion: { in: INTERVALS }
 
   def recalculate_stats
-    ping_times = PingThing.for_date_range_and_network(
+    ping_times = PingThing.for_reported_at_range_and_network(
       network,
       interval.minutes.ago,
       Time.now
--- a/app/models/ping_thing_stat.rb
+++ b/app/models/ping_thing_stat.rb
@@ -44,7 +44,7 @@ class PingThingStat < ApplicationRecord
   end
 
   def recalculate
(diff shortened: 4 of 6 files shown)


test/factories/ping_thing_recent_stats.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :ping_thing_recent_stat do
+    interval { 5 }
```
radek-polcode, line 5: I'd change that to trait named interval_5 similar to interval_60

test/models/ping_thing_recent_stat_test.rb
```diff
@@ -0,0 +1,82 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+class PingThingRecentStatTest < ActiveSupport::TestCase
+  include ActionCable::TestHelper
+
+  setup do
```
radek-polcode, line 8: looks like it;'s not needed

test/models/ping_thing_recent_stat_test.rb
```diff
@@ -0,0 +1,82 @@
...
+  end
+
+  test "#to_builder returns required attributtes" do
+    ping_stat = create(:ping_thing_recent_stat)
+    required_keys = PingThingRecentStat::FIELDS_FOR_API.map(&:to_s)
+
+    assert_equal required_keys, ping_stat.to_builder.attributes!.keys
+  end
+
+  test "creating new record brodcasts message" do
```
radek-polcode, line 78: Fill or remove

test/models/ping_thing_recent_stat_test.rb
```diff
@@ -0,0 +1,82 @@
+# frozen_string_literal: true
+
+require "test_helper"
+
+class PingThingRecentStatTest < ActiveSupport::TestCase
+  include ActionCable::TestHelper
```
radek-polcode, line 6: Is it needed?

app/models/ping_thing.rb
```diff
@@ -39,7 +39,11 @@ class PingThing < ApplicationRecord
   validates_length_of :application, maximum: 80, allow_blank: true
   validates :network, inclusion: { in: %w(mainnet testnet) }
   validates :signature, length: { in: 64..128 }
- 
+
+  scope :for_date_range_and_network, -> (network, from, to) {
```
radek-polcode, line 43: Maybe it'd be better to use reported_at in the scope name so it's more descriptive like `for_reported_at_range_and_network`

radek-polcode (approved):

test/factories/ping_thing_recent_stats.rb
kbiala: This field is obligatory, otherwise ping stat won't be created. I can add a trait additionally, but we need to have interval value here as well.

test/models/ping_thing_recent_stat_test.rb
kbiala: Will be in a follow-up task, so I added it to remember to add a test as well.

PR #521, opened 2022-07-26
182816572_create_data_center_for_validators_without_ip

radek-polcode:
#### What's this PR do?
- Create script to append validators to data center with key Unknown when no ip exists

- Run: `d = DataCenter.create(traits_autonomous_system_number: 0, city_name: "Unknown")`, `d.data_center_hosts.create(host: nil)` (it is needed only for one time)
- Run script locally, you should see new Data Center "Unknown" with some validators assigned
- Check the log file

- [URL to the PivotalTracker ticket](https://www.pivotaltracker.com/story/show/182816572)

#### Things to watch out for


maciejkocylapc (changes requested):
You should exclude "unknown" data center from counting data center score.

script/data_centers_scripts/append_to_unknown_data_center.rb
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+# RAILS_ENV=production bundle exec ruby script/data_centers_scripts/append_to_unknown_data_center.rb
+require_relative '../../config/environment'
+log_path = Rails.root.join('log', 'append_to_unknown_data_center.log')
+@logger ||= Logger.new(log_path)
+
+@logger.info "Script has started at #{Time.now}"
+
+begin
```
maciejkocylapc, line 10: there's no need to use begin ... end without rescue

script/data_centers_scripts/assign_data_center_scores.rb
```diff
@@ -42,6 +42,9 @@
 
 @data_centers = @data_centers.sort_by { |_k, v| -v[:count] }
 @data_centers.each do |k, v|
+  # We don't want to update score for validators from from unknown data_center
```
maciejkocylapc, line 45:
```suggestion
  # We don't want to update score for validators from unknown data_center
```

app/models/data_center.rb
```diff
@@ -47,6 +47,9 @@
 #  index_data_centers_on_data_center_key  (data_center_key)
 #
 class DataCenter < ApplicationRecord
+  # Data Center for validators not assigned anywhere (mostly due to lack of validator_ip) 
+  UNKNOWN_DATA_CENTER_KEY = "0--Unknown"
```
maciejkocylapc, line 51: I would just call it unknown - it should look better in views

radek-polcode: It's for the ASN view to have number

maciejkocylapc (approved):

kbiala:
Creation of "Unknown" data center should be included somewhere for development, eg. in seeds

PR #577, opened 2022-10-05
183466933 add gossip node status

kbiala:
#### What's this PR do?
- Adds is_active attribute to gossip nodes

- Run migration
- Confirm the script runs without any errors: `rails r script/update_gossip_nodes.rb`
- If you had some gossip nodes already in your db, then those which don't exist anymore should become inactive (check the number in console `GossipNode.where(is_active: false).count`)
- If you didn't have any gossip nodes in your local db, then all new nodes should be active (confirm in console `GossipNode.where(is_active: false).count` - should be 0)
- Use API to get gossip nodes: https://stage.validators.app/api-documentation?locale=en&network=mainnet#gossip-nodes. Confirm that the output no longer includes inactive gossip nodes.

- [URL to the PivotalTracker ticket](https://www.pivotaltracker.com/n/projects/2177859/stories/183466933)

#### Things to watch out for

### DIFF
--- a/app/logic/gossip_node_logic.rb
+++ b/app/logic/gossip_node_logic.rb
@@ -43,15 +43,16 @@ module GossipNodeLogic
 
   def set_inactive_nodes_status
     lambda do |p|
-      active_nodes_identities = p.payload[:current_nodes].map { |cn| cn["identityPubkey"] }
+      active_nodes_identities = p.payload[:current_nodes]&.map { |cn| cn["identityPubkey"] }
       GossipNode.where(network: p.payload[:network]).each do |gn|
-        unless active_nodes_identities.include? gn.account
+        unless active_nodes_identities&.include? gn.account
           gn.is_active = false
-          gn.save if gn.changed?
+          gn.save if gn.will_save_change_to_attribute?(:is_active)
         end
       end
       Pipeline.new(200, p.payload)
     rescue StandardError => e
+      puts e
       Pipeline.new(500, p.payload, "Error from set_inactive_nodes_status", e)
     end
   end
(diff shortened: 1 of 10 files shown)


kamtarnowski (approved):

kamtarnowski:
I did a local tests, but have mention that I didn't have gossip nodes, which does not exist anymore (inactive).

So the following in my case, did return 0 value.
`GossipNode.where(is_active: false).count`

app/logic/gossip_node_logic.rb
```diff
@@ -40,6 +41,21 @@ def update_nodes
     end
   end
 
+  def set_inactive_nodes_status
+    lambda do |p|
+      active_nodes_identities = p.payload[:current_nodes]&.map { |cn| cn["identityPubkey"] }
+      GossipNode.where(network: p.payload[:network]).each do |gn|
```
maciejkocylapc, line 47: perhaps it Could work without running through each gn? I. e. You could do `GossipNode.where(network: p.payload[:network]).where.not(account: active_nodes_identities).update_all(is_active: false)`

test/logic/gossip_node_logic_test.rb
```diff
@@ -55,12 +55,25 @@ class GossipNodeLogicTest < ActiveSupport::TestCase
     end
   end
 
+  test "set_inactive_nodes_status correctly updates status for inactive nodes" do
+    not_existing_node = create(:gossip_node, account: "AAAAAAAAAA", network: @network)
+    assert_equal true, not_existing_node.is_active
```
maciejkocylapc, line 60:
```suggestion
    assert not_existing_node.is_active
```

maciejkocylapc (changes requested):
You could also add a test checking if newly created nodes are marked as active

maciejkocylapc (approved):

PR #592, opened 2022-10-17
183408113 active stake front component

maciejkocylapc:
#### What's this PR do?
- create cluster stat model with active stake, and display it on the front component

- run migrations
- run tests
- run daemons and sidekiq to check if new CLusterStat

-[ URL to the PivotalTracker ticket](https://www.pivotaltracker.com/story/show/183408113)

#### Things to watch out for


psuchta (approved):

app/javascript/packs/validators/circulating_supply.vue
```diff
@@ -10,9 +10,13 @@
           <strong class="text-success" v-if="circulating_supply">{{ circulating_supply }}&nbsp;SOL</strong>
         </div>
 
-        <small class="text-muted" v-if="total_circulating_supply">
```
kamtarnowski, line 13: Why removing this?

app/javascript/packs/validators/circulating_supply.vue
```diff
@@ -42,9 +47,26 @@
       this.update_circulating_supply();
       this.set_continuous_supply_update();
     },
+    mounted: function(){
+      this.$cable.subscribe({
+          channel: "SolPriceChannel",
```
kamtarnowski, line 52: Maybe it would be better to create ClusterStatChannel instead of using one that is responsible for other component (sol prices)?

app/views/cluster_stats/index.json.jbuilder
```diff
@@ -0,0 +1 @@
+json.total_active_stake @total_active_stake
```
kamtarnowski, line 1: Are we using it anywhere? If not please remove it.

daemons/front_stats_update_daemon.rb
```diff
@@ -2,15 +2,38 @@
...
+    testnet: ClusterStat.by_network("testnet").last&.attributes
+  }
+end
+
+@last_cluster_stats = get_latest_cluster_stats
+
 loop do
   begin
+    @counter += 1
+    if @counter == update_stats_time / sleep_time
```
kamtarnowski, line 22: I'm not getting the point. 
The variables update_stats_time and sleep_time are constant and will not change. So executing the above `update_stats_time / sleep_time` will return value of 12.

Can we just define `update_stats_time` only and compare it with the `@counter`?

maciejkocylapc: if sleep time changes we still want to have 1 minute between updates

app/javascript/packs/validators/circulating_supply.vue
maciejkocylapc: I don't want to create too many channels, but I renamed the one for more general and fitting title

kamtarnowski (approved):
LGTM

PR #606, opened 2022-10-21
183640495 fix error with duplicated keys for a map

kbiala:
#### What's this PR do?
- Fixes an error with duplicated keys

- It's difficult to reproduce this error locally. I would advise to test it on staging, and compare with production version.
- Error description: on production there's no bubble for Germany (that's because 'DE' key is already taken by Delaware, US). After deploy to staging, this problem should be resolved, and you should be able to see a bubble for Germany. 

- [URL to the PivotalTracker ticket](https://www.pivotaltracker.com/n/projects/2177859/stories/183640495)

#### Things to watch out for


kamtarnowski:
It turned out that above behavior is correct.

PR #617, opened 2022-10-27
183639579 - reduce leader component intervals

kamtarnowski:
#### What's this PR do?
PR reduces the interval for fetching data of leaders component.

* restart the `daemons/front_stats_update_daemon.rb`
* run the `daemons/leader_stats_update_daemon.rb`

Verify if the homepage is updated correctly and includes a valid data.

[- URL to the PivotalTracker ticket](https://www.pivotaltracker.com/story/show/183639579)

#### Things to watch out for

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -120,6 +120,7 @@ namespace :deamons do
           execute :systemctl, '--user', :start, :gather_vote_account_details
           execute :systemctl, '--user', :start, :process_ping_thing
           execute :systemctl, '--user', :start, :front_stats_update
+          execute :systemctl, '--user', :start, :leader_stats_update
         end
       end
     end
@@ -137,6 +138,7 @@ namespace :deamons do
           execute :systemctl, '--user', :stop, :gather_vote_account_details
           execute :systemctl, '--user', :stop, :process_ping_thing
           execute :systemctl, '--user', :stop, :front_stats_update
+          execute :systemctl, '--user', :stop, :leader_stats_update
         end
       end
     end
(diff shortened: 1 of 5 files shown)


daemons/leader_stats_update_daemon.rb
```diff
@@ -0,0 +1,51 @@
...
+
+require_relative "../config/environment"
+
+@leaders_limit = 6
+@fetched_leaders_limit = 20
+@mainnet = "mainnet"
+@networks = [@mainnet, "testnet"].freeze
+sleep_time = 2 # seconds
+
+def all_leaders
```
maciejkocylapc, line 11: perhaps it would be a good idea to move all the methods to separate module in concerns? It would be a bit less messy.

config/deploy.rb
```diff
@@ -154,6 +156,7 @@
           execute :systemctl, '--user', :restart, :gather_vote_account_details
           execute :systemctl, '--user', :restart, :process_ping_thing
           execute :systemctl, '--user', :restart, :front_stats_update
+          execute :systemctl, '--user', :restart, :leader_stats_update
```
kbiala, line 159: Please comment this line out. This will break the deploy. It must be done in the next step, after the script is already on the server.

kbiala (changes requested):

daemons/leader_stats_update_daemon.rb
kamtarnowski: Sure, I've placed this logic in daemons/concerns.

config/deploy.rb
kamtarnowski: Thanks for warning me. I've commented this for restart block.

maciejkocylapc (approved):

PR #629, opened 2022-11-07
183727022 - fix method determining is validator a delinquent

kamtarnowski:
#### What's this PR do?
PR fixes the logic concerning validator activation in the service.

On localhost the following should be performed within the console: `ValidatorCheckActiveService.new.update_validator_activity`

On stage the cron should run the `ValidatorCheckActiveWorker`.

In both cases this validator https://www.validators.app/validators/E2Ydx12gumhPgxcFfxmaz796aC5it3vooZvAJYaaaEkz?locale=en&network=testnet should return status active (if is being voting).

[- URL to the PivotalTracker ticket](https://www.pivotaltracker.com/story/show/183727022)

#### Things to watch out for


app/services/validator_check_active_service.rb
```diff
@@ -86,16 +86,7 @@ def update_scorable(validator)
...
-                                              account: validator.account,
-                                              delinquent: false
-                                            )
-                                            .where(
-                                              'created_at > ?',
-                                              DateTime.now - @delinquent_time
-                                            )
-
-    nondelinquent_history.exists?
+    !validator.delinquent?
```
maciejkocylapc, line 89: I would merge the two options like `nondelinquent_history.exists? || !validator.delinquent?`

test/services/validator_check_active_service_test.rb
```diff
@@ -67,16 +67,14 @@ class ValidatorCheckActiveWorkerTest < ActiveSupport::TestCase
     assert v.reload.is_rpc
   end
 
-  test 'validator delinquent for too long should be inactive' do
```
maciejkocylapc, line 70: I don't think this test should be deleted

maciejkocylapc (changes requested):

app/services/validator_check_active_service.rb
```diff
@@ -110,4 +102,8 @@ def acceptable_stake?(validator)
     with_acceptable_stake.exists?
   end
 
+  def delinquent_history(validator)
```
maciejkocylapc, line 105: this method is not the same as the previous one and does not work as it should

maciejkocylapc (changes requested):

app/services/validator_check_active_service.rb
kamtarnowski: Fixed.

maciejkocylapc (approved):

PR #636, opened 2022-11-09
183749738-handle-unknown-dc-on-leader-map

maciejkocylapc:
#### What's this PR do?
- do not display leaders with 0-unknown data center on a map

- check the map - perhaps on staging it should be possible to wait for a leader with unknown dc - it should not be displayed on a map

- [URL to the PivotalTracker ticket](https://www.pivotaltracker.com/story/show/183749738)

#### Things to watch out for

### DIFF
--- a/app/javascript/packs/validators/validators_map.vue
+++ b/app/javascript/packs/validators/validators_map.vue
@@ -30,7 +30,7 @@
           </span>
         </div>
 
-        <div v-if="current_leader"
+        <div v-if="leader_valid"
              :style="{ left: position_horizontal(current_leader.location_longitude),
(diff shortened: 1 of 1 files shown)


app/javascript/packs/validators/validators_map.vue
```diff
@@ -109,9 +109,14 @@
     },
     watch: {
     },
-    computed: mapGetters([
-      'network'
-    ]),
+    computed: {
+      leader_valid(){
```
kamtarnowski, line 113: For methods which returns a boolean I would suggest the following `is_leader_valid`.

If it was ruby, we would use the `leader_valid?`.

kamtarnowski (approved):

PR #637, opened 2022-11-09
183730524 - add epoch pythnet to sol price

kamtarnowski:
#### What's this PR do?
PR adds epoch_pythnet integer.

Just execute `rails db:migrate` and check the schema.

[- URL to the PivotalTracker ticket](https://www.pivotaltracker.com/story/show/183730524)

#### Things to watch out for

### DIFF
--- a/db/migrate/20221109120726_add_epoch_pythnet_to_sol_price.rb
+++ b/db/migrate/20221109120726_add_epoch_pythnet_to_sol_price.rb
@@ -1,0 +1,5 @@
+class AddEpochPythnetToSolPrice < ActiveRecord::Migration[6.1]
+  def change
+    add_column :sol_prices, :epoch_pythnet, :integer
+  end
+end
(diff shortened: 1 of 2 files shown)


db/migrate/20221109120726_add_epoch_pythnet_to_sol_price.rb
```diff
@@ -0,0 +1,5 @@
+class AddEpochPythnetToSolPrice < ActiveRecord::Migration[6.1]
+  def change
+    add_column :sol_prices, :epoch_pythnet, :integer
```
psuchta, line 3: We agreed that we don't need to add `epoch_pythnet` to the `sol_prices` table as it is not used in any view

psuchta (changes requested):

db/migrate/20221109120726_add_epoch_pythnet_to_sol_price.rb
kamtarnowski: Recently, It turned out as above.

PR #645, opened 2022-11-14
183730467 - add pythnet to scripts

psuchta:
#### What's this PR do?
- Add pythnet to scripts

- run `rails r script/add_current_epoch.rb`

- [URL to the PivotalTracker ticket](https://www.pivotaltracker.com/story/show/183730467)

#### Things to watch out for

### DIFF
--- a/app/services/solana_rpc_client.rb
+++ b/app/services/solana_rpc_client.rb
@@ -39,4 +39,14 @@ class SolanaRpcClient
       cluster: Rails.application.credentials.solana[:testnet_urls].first
     )
   end
+
+  def network_client(network)
+    network_cluster = Rails.application.credentials.solana["#{network}_urls".to_sym].first
+    # Reset the client when the network is different from the one already initialized
+    if @cluster != network_cluster
+      @cluster = network_cluster
+      @client = nil
+    end
+    client
+  end
 end
--- a/app/services/solana_rpc_client.rb
+++ b/app/services/solana_rpc_client.rb
@@ -7,7 +7,6 @@ class SolanaRpcClient
 
   # Usage:
   # client  = SolanaRpcClient.new(cluster: 'custom_cluster') # your custom cluster
-  # default = SolanaRpcClient.new.default_client # uses cluster from initializers/solana_rpc_ruby
   # mainnet = SolanaRpcClient.new.network_client('mainnet') # uses cluster specific for a passed network.
   #           network_client method can receive 'mainnet','testnet' or 'pythnet' network name
   #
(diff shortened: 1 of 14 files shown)


app/services/solana_rpc_client.rb
```diff
@@ -39,4 +39,14 @@ def testnet_client
       cluster: Rails.application.credentials.solana[:testnet_urls].first
     )
   end
+
+  def network_client(network)
```
maciejkocylapc, line 43: I would add a comment to two methods above that they're deprecated - unless You're certain that they're not used anywhere and it's safe to remove them

app/services/solana_rpc_client.rb
```diff
@@ -1,13 +1,15 @@
 # frozen_string_literal: true
```
kamtarnowski, line 1: If we are cleaning this class up we could remove the `default_client` method. So please make it as well.

app/services/solana_rpc_client.rb
```diff
@@ -1,13 +1,15 @@
...
 
 class SolanaRpcClient
+  class InvalidNetwork < StandardError; end
+
   attr_reader :cluster
 
   # Usage:
   # client  = SolanaRpcClient.new(cluster: 'custom_cluster') # your custom cluster
   # default = SolanaRpcClient.new.default_client # uses cluster from initializers/solana_rpc_ruby
```
kamtarnowski, line 10: If removing `default_client` this also should be deleted.

test/services/solana_rpc_client_test.rb
```diff
@@ -24,4 +24,25 @@ class SolanaRpcClientTest < ActiveSupport::TestCase
...
+  test "#network_client returns correct client" do
+    # No token, for privacy reason.
+    vcr_cassette(@namespace, __method__) do
+      rpc_client = SolanaRpcClient.new
+      mainnet = rpc_client.network_client("mainnet")
+
+      # returns same client when same network used again
+      assert_equal rpc_client.network_client("mainnet"), mainnet
+      # returns different client when different network used
+      refute rpc_client.network_client("pythnet") == mainnet
```
kamtarnowski, line 37: When we did `rpc_client.network_client("pythnet")` here, then the client will be `nil`. Should it be like this?

psuchta: rpc_client.network_client("pythnet") returns `SolanaRpcRuby::MethodsWrapper` object, which doesnt have `client` method/attribute
If you run
```
rpc = SolanaRpcClient.new
rpc.network_client("pythnet")
rpc.client
```
It will return correct wrapper

kamtarnowski (approved):

PR #661, opened 2022-11-23
183769802 - use find_each on large ActiveRecord relations

kamtarnowski:
#### What's this PR do?
PR adds find_each method for large ActiveRecord relations. Using it instead of each method.

- `rails test`

Changes impact on the following tasks, so please run them to make sure they work fully correct:
- `daemons/gather_rpc_mainnet_daemon.rb`
- `daemons/gather_vote_account_details_daemon.rb`
- perform in console: `ValidatorCheckActiveService.new.update_validator_activity`
- `script/validator_ips/set_active_ip_for_gossip_nodes.rb`
- `script/one_time_scripts/back_fill_commission_history.rb`
- `script/one_time_scripts/set_is_active_in_validator_ips.rb`

[- URL to the PivotalTracker ticket](https://www.pivotaltracker.com/story/show/183769802)

#### Things to watch out for


psuchta (changes requested):

app/logic/report_logic.rb
```diff
@@ -19,7 +19,7 @@ def build_skipped_slot_percent
 
       # Create a results array and insert the data
       result = []
-      block_history.each do |history|
+      block_history.find_each do |history|
```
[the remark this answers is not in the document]
kamtarnowski: Yes, you are right. I will make a fix.

app/services/gatherers/vote_account_details_service.rb
```diff
@@ -10,14 +10,13 @@ def initialize(network:, config_urls:, always_update: false)
       @always_update = always_update
       @network = network
       @config_urls = config_urls
-      @range_start = VoteAccount.where(network: @network).order(id: :asc).first.id
-      @range_end = VoteAccount.where(network: @network).order(id: :asc).last.id
+      @range_start = VoteAccount.where(network: @network).first.id
```
psuchta, line 13: Is it possible that VoteAccounts are not ordered as we expect here?
For example (ids) :  1,2,3,5,6,4
Then @range_start will be 1 and @range_end will be 4. Some VoteAccounts will be skipped in that case.

To be honest I don't see a reason for specifying `where(id: @range_start..@range_end`  down below. I think we can delete that for where clause

psuchta (changes requested):

app/services/gatherers/vote_account_details_service.rb
kamtarnowski: I agree, this range logic is useless here. I did remove it.

psuchta (approved):

PR #674, opened 2022-11-30
183882661 - enable turbolinks on sol prices page

kamtarnowski:
#### What's this PR do?
Unfortunately we cannot load vue components on real `turbolinks:load` events. These components are being loaded on `DOMContentLoaded` event instead. 

Notice: if turbolinks data is disabled (at our homepage e.g.) then turbolinks reloads all data, so the events `turbolinks:load` and `DOMContentLoaded` do perform at the same point.

This PR only turns on turbolinks on Sol Prices page.

- check sol prices page if the content is loaded
- at Sol Prices page move to page where turbolinks are activated
- click "back button" and check whether the `<body>` tag has only been changed (`<head>` should stay untouched)

[- URL to the PivotalTracker ticket](https://www.pivotaltracker.com/story/show/183882661)

#### Things to watch out for

### DIFF
--- a/app/views/layouts/_footer.html.erb
+++ b/app/views/layouts/_footer.html.erb
@@ -29,7 +29,7 @@
         <li class="nav-item" data-turbolinks="false">
           <%= link_to 'Ping Thing', ping_things_path(network: params[:network]), class: 'nav-link' %>
         </li>
-        <li class="nav-item" data-turbolinks="true">
+        <li class="nav-item">
           <%= link_to 'Sol Prices', sol_prices_url(exchange: SolPrice::EXCHANGES[0], filtering: 7), class: 'nav-link' %>
         </li>
         <li class="nav-item">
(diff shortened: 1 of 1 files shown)


app/views/layouts/_footer.html.erb
```diff
@@ -29,7 +29,7 @@
         <li class="nav-item" data-turbolinks="false">
           <%= link_to 'Ping Thing', ping_things_path(network: params[:network]), class: 'nav-link' %>
         </li>
-        <li class="nav-item" data-turbolinks="false">
+        <li class="nav-item" data-turbolinks="true">
```
maciejkocylapc, line 32: you can delete data-turbolinks as "true" is the default behavior

kamtarnowski: Yeah, you have totally right.

PR #675, opened 2022-11-30
183792092 merge leaders with map

kbiala:
#### What's this PR do?
- Moves leaders component to map

- On localhost run leaders daemon for selected network (eg. `rails r daemons/leader_stats_mainnet_update_daemon.rb`)
- Go to homepage and check the map

- [URL to the PivotalTracker ticket
](https://www.pivotaltracker.com/story/show/183792092)

#### Things to watch out for


app/javascript/packs/validators/validators_map_leaders.js
```diff
@@ -0,0 +1,66 @@
+import Vue from 'vue/dist/vue.esm'
+
+var ValidatorsMapLeaders = Vue.component('ValidatorsMapLeaders', {
+    props: {
```
kamtarnowski, line 4: Is there something wrong with the indentation?

app/views/public/home.html.erb
```diff
@@ -15,11 +14,14 @@
   <validators-map network="<%= @network %>"/>
 </div>
 
+<!--
```
kamtarnowski, line 17: Can we remove these unused lines of code?

kamtarnowski (changes requested):
Please also remove 
`<%= javascript_pack_tag 'leaders/leaders_stats_component.js' %>`
from home_new.html.erb

kamtarnowski:
The code looks pretty good, also the homepage seems much better.

kbiala:
@kamtarnowski done

kamtarnowski (approved):

kamtarnowski:
That is not a blocker, but maybe we could just remove this view - `app/views/public/home_new.html.erb`, to not maintain both similar views. Any reference to `home_new` path would be defined in the routes file.

kbiala:
> That is not a blocker, but maybe we could just remove this view - `app/views/public/home_new.html.erb`, to not maintain both similar views. Any reference to `home_new` path would be defined in the routes file.

Very bad idea - this view was supposed to be different, to show things that are in progress and pending for acceptance.
Anyway, we probably don't need it at all anymore. But in any case it's completely unrelated to this story.

PR #688, opened 2022-12-15
183982396 - track authorities for vote account

kamtarnowski:
#### What's this PR do?
PR adds versioning of `authorized_withdrawer` in vote accounts. It also appends new attribute - `authorized_voters`

Please test using the `rails c`.

After any changes made to either `authorized_withdrawer` or `authorized_voters` the change history should be accessible via `account_authority_histories` table. Please verify it.

[- URL to the PivotalTracker ticket](https://www.pivotaltracker.com/story/show/183982396)

#### Things to watch out for


app/models/vote_account.rb
```diff
@@ -61,4 +62,12 @@ def to_builder
       va.vote_account self.account
     end
   end
+
+  def create_vote_account_history
+    vote_account_histories.create(
```
kbiala, line 67: 1. You cannot do it this way.
First of all, VoteAccountHistories store historic state of vote account, which is used in multiple places accross the app and it cannot be empty (or empty except those two fields that you added)! It breaks many things, account details view to just give one example.
Secondly, it's a really bad idea to mix up two data formats in the same model. If we wanted to store this data in VoteAccountHistory model, then these two values would be the same attribute (authorized_withdrawer) in two different records. And it should be moved to the pipeline (as I wrote in the story).

But we might need to create a separate table for these values after all (see story description), because we prune VoteAccountHistory table after 60 days. And Brian most probably will want to keep these data much longer.

2. See story description - we want to track all authorities, not only withdrawer authority.

3. You'll also probably need to modify and adjust to your changes VoteAccountDetailsService service (the part related to withdrawer authority).

kbiala (changes requested):

kamtarnowski:
Zrzut ekranu 2022-12-20 o 11 02 02 (3)](https://user-images.githubusercontent.com/7115768/208640023-b66a6caa-2598-45f4-a9d4-36e22eeb5521.png)
g)

This detail view is still not valid.

kbiala (approved):

PR #710, opened 2023-01-18
184004267 commission tracking script

maciejkocylapc:
#### What's this PR do?
- check get_inflation_reward endpoint for the commission changes we've not caught before
- run the check an hour after new epoch was created
- add missing commission changes

- run tests
- populate db with current and previous EpochWallClock
- try running the service in console
- check if new CommissionHistory records are created (it should have from_inflation_rewards: true)

- [URL to the PivotalTracker ticket](https://www.pivotaltracker.com/story/show/184004267)

#### Things to watch out for


app/services/track_commission_changes_service.rb
```diff
@@ -0,0 +1,74 @@
...
+    @previous_batch = Batch.where("created_at < ?", @current_batch.created_at)
+                           .where(network: network)
+                           .where.not(scored_at: nil)
+                           .last
+    @network = network
+    @solana_url = solana_url || NETWORK_URLS[@network]
+  end
+
+  def call
+    VoteAccount.includes(:validator).where(network: @network).in_batches(of: 100) do |batch|
```
kamtarnowski, line 17: `batch` in that case seems misleading. Maybe just `va` / `vote_account`

app/services/track_commission_changes_service.rb
```diff
@@ -0,0 +1,74 @@
...
+      accounts = batch.pluck(:account)
+      result = get_inflation_rewards(accounts)
+
+      batch.each_with_index do |va, idx|  
+        next unless result[idx] # some accounts have no inflation reward e. g. if it was inactive
```
kamtarnowski, line 22: You can use `compact_blank` on the active record relation instead.

kamtarnowski (changes requested):

app/services/track_commission_changes_service.rb
```diff
@@ -0,0 +1,74 @@
+# frozen_string_literal: true
+
+class TrackCommissionChangesService
+  include SolanaLogic
+
+  def initialize(current_batch: , network: "mainnet", solana_url: nil)
```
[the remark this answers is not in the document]
maciejkocylapc: It's optional, but I thought it would be more elastic and future-proof this way

app/services/track_commission_changes_service.rb
maciejkocylapc: Why is it misleading? it is a batch of records

app/services/track_commission_changes_service.rb
maciejkocylapc: I don't want to do that, because then my result count would be different than my batch count and I wouldn't be able to assign results to vote accounts

app/services/track_commission_changes_service.rb
```diff
@@ -0,0 +1,74 @@
...
+    ) do |commission|
+      commission.commission_after = validator_commission
+      commission.epoch_completion = 0.05
+      commission.batch_uuid = @current_batch.uuid
+      commission.from_inflation_rewards = true
+    end
+  end
+
+  def check_commission_for_previous_epoch(va, cli_rewards)
+    validator_commission = va.validator.commission
```
kbiala, line 51: Isn't `va.validator.commission` the value of current commission rate? It should be the previous value here, pulled from Solana get_inflation_rewards method.

kbiala (changes requested):

app/services/track_commission_changes_service.rb
```diff
@@ -0,0 +1,94 @@
...
+    VoteAccount.includes(:validator).where(network: @network).in_batches(of: 100) do |va_batch|
+      accounts = va_batch.pluck(:account)
+      result = get_inflation_rewards(accounts)
+
+      va_batch.each_with_index do |va, idx|  
+        next unless result[idx] # some accounts have no inflation reward e. g. if it was inactive
+  
+        # check if commission from db matches commission from get_inflation_reward
+        if va.validator.commission != result[idx]["commission"].to_f
+          create_commission_for_current_epoch(va, result[idx])
```
kbiala, line 26: Let's change the order, previous first.

app/services/track_commission_changes_service.rb
```diff
@@ -0,0 +1,94 @@
...
+        from_inflation_rewards: true,
+        created_at: @current_batch.created_at
+      )
+    end
+  end
+
+  def create_commission_for_previous_epoch(va, cli_rewards)
+    validator_commission = va.validator.commission
+
+    first_commission_from_epoch = va.validator.commission_histories.where(
```
kbiala, line 63: We should compare to the last commission, not first.

app/services/track_commission_changes_service.rb
```diff
@@ -0,0 +1,94 @@
...
+
+  def create_commission_for_current_epoch(va, cli_rewards)
+    validator_commission = va.validator.commission
+
+    last_commission_from_epoch = va.validator.commission_histories.where(
+      epoch: cli_rewards["epoch"] + 1,
+      network: @network,
+    ).last
+
+    unless last_commission_from_epoch && \
```
kbiala, line 43: We should compare to the first commission

kbiala (changes requested):

kbiala (approved):

PR #722, opened 2023-02-07
Filter validators by admin_warning if present

kbiala:
#### What's this PR do?
- Adds optiona admin_warning param 

- Generate some admin_warnings for testing, eg. `Validator.scorable.first(25).each { |v| v.update admin_warning: 'test' } `
- Run the server, go to validators page by a link in top menu. Confirm that default list is unchanged (compared to master branch).
- Now add `&admin_warning=ASD` param to URL at the end (so full URL would look for example like this `http://localhost:3000/validators?locale=en&network=mainnet&admin_warning=true`). Confirm that now you can only see validators with admin_warnings that you edited in point 1.
- Replace param with `&admin_warning=false`. Confirm that now only validators without admin_warnings are listed.
- Confirm that incorrect parameters are ignored, eg. `&admin_warning=ASD`
- Combine with search feature and compare results (you will add admin_warning by hand, it's meant to be hidden)

#### Things to watch out for

### DIFF
--- a/app/controllers/validators_controller.rb
+++ b/app/controllers/validators_controller.rb
@@ -21,7 +21,8 @@ class ValidatorsController < ApplicationController
       sort_order: validators_params[:order],
       limit: @per,
       page: validators_params[:page],
-      query: validators_params[:q]
+      query: validators_params[:q],
+      admin_warning: validators_params[:admin_warning]
     )
 
     @batch = Batch.last_scored(validators_params[:network])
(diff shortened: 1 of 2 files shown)


test/queries/validator_query_test.rb
```diff
@@ -64,10 +64,45 @@ class ValidatorQueryTest < ActiveSupport::TestCase
...
+    result = ValidatorQuery.new.call(network: @testnet_network, admin_warning: admin_warning)
+
+    assert_equal 1, result.count
+    assert_equal [@testnet_network], result.pluck(:network).uniq
+    assert_equal ["test warning"],  result.pluck(:admin_warning).uniq
+
+    admin_warning = "false"
+    result = ValidatorQuery.new.call(network: @testnet_network, admin_warning: admin_warning)
+
+    assert_equal (Validator.where(network: @testnet_network).count - 1), result.count
```
maciejkocylapc, line 81: you could check if the same ids are returned

PR #735, opened 2023-02-24
184544114 - add trent-mode page

kamtarnowski:
#### What's this PR do?
PR adds new trent-mode page.

- Enter the `/trent-mode` page. Verify its content. Use search bar and confirm that it works. Switch networks to confirm that they're correctly passed to search query.
- check the `/validators` page and verify that there is no negative impact.

[- URL to the PivotalTracker ticket](https://www.pivotaltracker.com/story/show/184544114)

#### Things to watch out for


app/controllers/concerns/validators_controller_helper.rb
```diff
@@ -56,4 +56,51 @@ def set_boolean_field(value)
...
+
+  def search_with_results_action
+    @per = 25
+
+    if validators_params[:watchlist] && !current_user
+      flash[:warning] = "You need to create an account first."
+      redirect_to new_user_registration_path and return
+    end
+
+    watchlist_user = validators_params[:watchlist] ? current_user&.id : nil
```
kbiala, line 68: The whole logic around watchlist etc. is not a part of task requirements.

app/views/public/trent_mode.html.erb
```diff
@@ -0,0 +1,34 @@
+<h1>Trent Mode</h1>
```
kbiala, line 1: See other pages for a proper html+css structure that we use.

app/helpers/validators_helper.rb
```diff
@@ -142,4 +142,12 @@ def validator_score_attrs(validator)
       ]
     )
   end
+
+  def validators_search_page_attrs(trent_mode)
```
kbiala, line 146: Missing default value

kbiala (changes requested):
The solution is inconsistent with the requirements described in task description:
> [page] that includes only the search bar & filtered validator search results.

kbiala (changes requested):
Also, I don't see a link on top of the page, that was also required in this story.

app/controllers/public_controller.rb
```diff
@@ -51,12 +53,20 @@ def commission_histories
     end
   end
 
+  def trent_mode
```
kbiala, line 56: This method is related to validators, so it should be placed in Validators Controller, not in Public Controller

kbiala (changes requested):

app/controllers/validators_controller.rb
```diff
@@ -25,14 +25,7 @@ def index
...
 
-    @batch = Batch.last_scored(validators_params[:network])
-
-    if @batch
-      @this_epoch = EpochHistory.where(
-        network: validators_params[:network],
-        batch_uuid: @batch.uuid
-      ).first
-    end
+    set_batch_and_epoch
```
maciejkocylapc, line 28: perhaps this should be a before_action ?

app/views/layouts/_header.html.erb
```diff
@@ -23,6 +23,9 @@
           <li class="nav-item">
             <%= link_to "Getting Started", faq_path, class: 'nav-link', "data-turbolinks": "false" %>
           </li>
+          <li class="nav-item">
+            <%= link_to "Trent Mode", trent_mode_path(network: params[:network]), class: 'nav-link', "data-turbolinks": "false" %>
```
maciejkocylapc, line 27: do we have to turn off turbolinks here ?

app/controllers/validators_controller.rb
kbiala: It could be. But on trent page, when there's no :q param, I don't want to query db for batch epoch unnecessarily. I'll set it as before_action on index only.

app/views/layouts/_header.html.erb
kbiala: We don't. Good catch

PR #738, opened 2023-02-27
184544324 filter by jito

maciejkocylapc:
#### What's this PR do?
- adds script that connects with the JITO api and assigns correct flag to validators in our db
- adds filter on /validators page displaying only jito validators

- run migrations
- run tests
- run script `rails r script/update_jito_validators.rb` and check if the flag is updated
- see if the filter works properly on /validators page

- [URL to the PivotalTracker ticket](https://www.pivotaltracker.com/story/show/184544324)

#### Things to watch out for


kamtarnowski (changes requested):

script/update_jito_validators.rb
```diff
@@ -0,0 +1,24 @@
...
+
+uri = URI("https://kobe.mainnet.jito.network/api/v1/validators")
+response = Net::HTTP.get_response(uri)
+if response.is_a?(Net::HTTPSuccess)
+  json_response = JSON.parse(response.body)
+  jito_vote_accounts = json_response["validators"].map do |jito_validator|
+    jito_validator["vote_account"] if jito_validator["running_jito"]
+  end
+
+  jito_db_validators = Validator.joins(:vote_accounts)
```
kbiala, line 16: We should filter by active vote accounts only. 
Otherwise, we will also pull validators that used to be assocciated with these VoteAccounts in the past, but are using different vote accounts now.
And I would also add a test for this case.

app/controllers/validators_controller.rb
```diff
@@ -68,6 +69,6 @@ def at_33_stake_index(validators, batch, per_page)
   end
 
   def validators_params
-    params.permit(:watchlist, :network, :q, :page, :order, :admin_warning)
+    params.permit(:watchlist, :network, :q, :page, :order, :admin_warning, :jito)
```
kbiala, line 72: Let's call this method jito_collaboration as well. To avoid a conflict in the future between this feature, and the one that pulls a list of validators from Jito Pool (currently WIP).

kbiala (changes requested):

kbiala:
@maciejkocylapc Please take a look at conflicts.

script/update_jito_validators.rb
```diff
@@ -0,0 +1,24 @@
...
+                                .where(
+                                  "validators.network = ? AND vote_accounts.account IN (?)",
+                                  "mainnet",
+                                  jito_vote_accounts
+                                )
+  jito_db_validators.update_all(jito_collaboration: true)
+  Validator.where.not(id: jito_db_validators.pluck(:id)).update_all(jito_collaboration: false)
```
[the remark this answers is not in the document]
maciejkocylapc: this isn't correct

kbiala (approved):

kbiala (approved):

PR #741, opened 2023-03-03
184581886-fix-roi-for-pythnet

maciejkocylapc:
#### What's this PR do?
- allow 0 stakes to be taken into account for pythnet

- test on staging (check ROI for pythnet)

- [URL to the PivotalTracker ticket](https://www.pivotaltracker.com/story/show/184581886)

#### Things to watch out for

### DIFF
--- a/app/services/total_rewards_update_service.rb
+++ b/app/services/total_rewards_update_service.rb
@@ -14,7 +14,7 @@ class TotalRewardsUpdateService
     if completed_epoch
       vote_rewards = total_rewards_from_vote_accounts(completed_epoch)
       stake_rewards = total_rewards_from_stake_accounts(completed_epoch)
-      if vote_rewards > 0 && stake_rewards > 0
+      if (vote_rewards > 0 && stake_rewards > 0) || @network == "pythnet"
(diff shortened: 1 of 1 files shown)


app/services/total_rewards_update_service.rb
```diff
@@ -14,7 +14,7 @@ def call
     if completed_epoch
       vote_rewards = total_rewards_from_vote_accounts(completed_epoch)
       stake_rewards = total_rewards_from_stake_accounts(completed_epoch)
-      if vote_rewards > 0 && stake_rewards > 0
+      if (vote_rewards > 0 && stake_rewards > 0) || @network == "pythnet"
```
kamtarnowski, line 17: What about removing `if (vote_rewards > 0 && stake_rewards > 0) || @network == "pythnet"` condition and just making an update every time?

maciejkocylapc: on mainnet if the rewards are 0 that means someting is wrong (like bad rpc call) and it should never be 0

kbiala (approved):

PR #749, opened 2023-03-10
tweak-ping-thing-performance

maciejkocylapc:
#### What's this PR do?

- List the testing steps here

#### Things to watch out for

### DIFF
--- a/app/models/ping_thing.rb
+++ b/app/models/ping_thing.rb
@@ -69,18 +69,18 @@ class PingThing < ApplicationRecord
   end
 
   def broadcast
-    start = Time.now
     hash = {}
     hash.merge!(self.to_builder.attributes!)
     hash.merge!(self.user.to_builder.attributes!)
 
     ActionCable.server.broadcast("ping_thing_channel", hash)
-    Rails.logger.warn("ProcessPingThing: broadcast (#{self.id}): #{Time.now - start}s")
   end
 
   def update_stats_if_present
     start = Time.now
+    Rails.logger.warn("ProcessPingThing: update_stats_if_present (#{self.id}) start")
     stats = PingThingStat.by_network(network).between_time_range(reported_at)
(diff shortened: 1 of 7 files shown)


app/models/ping_thing.rb
```diff
@@ -79,8 +77,12 @@ def broadcast
   end
 
   def update_stats_if_present
+    start = Time.now
+    Rails.logger.warn("ProcessPingThing: update_stats_if_present (#{self.id}) start")
     stats = PingThingStat.by_network(network).between_time_range(reported_at)
+    Rails.logger.warn("ProcessPingThing: update_stats_if_present (#{self.id}) stats_count: #{stats.count}")
```
kbiala, line 83:
```suggestion
    Rails.logger.warn("ProcessPingThing: update_stats_if_present (#{self.id}) stats_count: #{stats&.count}")
```

app/services/process_ping_things_service.rb
```diff
@@ -16,9 +16,18 @@ def call
...
       )
+      ping_thing = PingThing.create(params)
+
+      # create sidekiq job to update stats if it's not a latest data
+      if raw.should_recalculate_stats_after_processing?
+        RecalculatePingThingStatsWorker.perform_async({
+          reported_at: ping_thing.reported_at,
+          network: ping_thing.network
+        })
+      end
```
kbiala, line 27: All good, except this doesn't tell us if `create` succedeed or not (ping_thing object will always be created, no matter if db record was created or not). And we should only create the worker if PingThing db record was created (eg. all required params like signature etc. are present and valid).
```suggestion
      ping_thing = PingThing.new(params)

      # create sidekiq job to update stats if it's not a latest data
      if ping_thing.save && raw.should_recalculate_stats_after_processing?
        RecalculatePingThingStatsWorker.perform_async({
          reported_at: ping_thing.reported_at,
          network: ping_thing.network
        })
      end
```

kbiala (changes requested):

kbiala (approved):

PR #750, opened 2023-03-10
update logic for count_average_validators_commission

kamtarnowski:
#### What's this PR do?

- List the testing steps here

[- URL to the PivotalTracker ticket](https://www.pivotaltracker.com/story/show/180982568)

#### Things to watch out for

### DIFF
--- a/app/logic/stake_logic.rb
+++ b/app/logic/stake_logic.rb
@@ -198,10 +198,24 @@ module StakeLogic
 
       StakePool.where(network: p.payload[:network]).each do |pool|
         validator_ids = pool.stake_accounts.active.pluck(:validator_id)
-        average_commission = ValidatorScoreV1.where(validator_id: validator_ids)
-                                             .average(:commission)
+        total_active_stake = 0
+        commissions = []
 
-        pool.update_attribute(:average_validators_commission, average_commission)
+        Validator.where(id: validator_ids)
+                 .includes(:stake_accounts, :validator_score_v1)
+                 .find_each do |validator|
+          validator_active_stake = validator.stake_accounts.active.where(
+            stake_pool: pool
+          ).sum(:active_stake)
+
+          total_active_stake += validator_active_stake
+
+          commissions.push(validator_active_stake * validator.score.commission)
+        end
+
+        average_commission = (commissions.sum / total_active_stake.to_f).round(2)
+
+        pool.update(average_validators_commission: average_commission)
       end
 
       Pipeline.new(200, p.payload)


app/models/validator.rb
```diff
@@ -81,6 +81,7 @@ class Validator < ApplicationRecord
 
   delegate :data_center_key, to: :data_center_host, prefix: :dch, allow_nil: true
   delegate :address, to: :validator_ip_active, prefix: :vip, allow_nil: true
+  delegate :commission, to: :validator_score_v1, allow_nil: true
```
kbiala, line 84: Commission is score's attribute, so there's no point of delegating it to score model.
Check score method content in validator model.

app/logic/stake_logic.rb
```diff
@@ -198,10 +198,24 @@ def count_average_validators_commission
...
       StakePool.where(network: p.payload[:network]).each do |pool|
         validator_ids = pool.stake_accounts.active.pluck(:validator_id)
-        average_commission = ValidatorScoreV1.where(validator_id: validator_ids)
-                                             .average(:commission)
+        total_active_stake = 0
+        commissions = []
+
+        Validator.where(id: validator_ids)
+                 .includes(:stake_accounts, :validator_score_v1)
+                 .find_each do |validator|
```
kbiala, line 206: I'd look for a way to refactor this. There's a lot of shared logic in this method and in `update_validator_stats` (pulling exactly the same validators from db, calculating total_active_stake again etc.).
Otherwise, we'll add another big query to db here.
Perhaps extract this logic to a separate method called before `update_validator_stats` and `count_average_validators_commission`.

kbiala (changes requested):

kbiala (changes requested):
Also, tests are missing

app/services/stake_pools/calculate_average_stats.rb
```diff
@@ -0,0 +1,85 @@
...
+      @pool = pool
+    end
+
+    def call
+      @sp_attrs = initial_attrs
+
+      Validator.where(id: validator_ids)
+               .includes(:stake_accounts, :validator_score_v1)
+               .find_each do |validator|
+        @validator = validator
```
maciejkocylapc, line 15: I don't think this is used anywhere

app/services/stake_pools/calculate_average_stats.rb
```diff
@@ -0,0 +1,85 @@
...
+    def uptime
+      (DateTime.now - last_delinquent.to_datetime).to_i
+    end
+
+    def lifetime
+      (DateTime.now - validator.created_at.to_datetime).to_i
+    end
+
+    def average_commission
+      (sp_attrs[:commissions].sum / sp_attrs[:total_active_stake].to_f).round(2)
```
maciejkocylapc, line 53: it's theoretically possible that total_active_stake == 0, so it should be rescued

app/services/stake_pools/calculate_average_stats.rb
kamtarnowski: I've added a getter in terms of validator in line 34.

maciejkocylapc (approved):

PR #765, opened 2023-03-28
184781969 sort versions by stake

maciejkocylapc:
#### What's this PR do?
- sort software versions by stake

- run tests
- check if software versions on home page are sorted by stake percent correctly

- [URL to the PivotalTracker ticket](https://www.pivotaltracker.com/story/show/184781969)

#### Things to watch out for


app/helpers/validators_helper.rb
```diff
@@ -75,7 +75,7 @@ def skipped_vote_percent(validator, batch)
   end
 
   def sort_software_versions(versions)
-    versions&.sort_by { |ver| Gem::Version.new(ver.keys.first)}&.reverse
+    versions&.sort_by { |ver| ver.values[0]["stake_percent"] }&.reverse
```
kamtarnowski, line 78: There are two negative scenarios: 
1) versions are invalid and AppSignal will send an alert (we would notice that faster and think about some fix)
In this case I would suggest to make a change 
```suggestion
    versions.sort_by { |ver| ver.values[0]["stake_percent"] }.reverse
```

2) versions are invalid (e.g. nil) and we do not Rails to raise any problem (software versions list will be blank on homepage)
```suggestion
    Array(versions).sort_by { |ver| ver.values[0]["stake_percent"] }&.reverse
```

kamtarnowski: IMO we should stick with the 1st scenario.

maciejkocylapc: I do not think this is a good idea. Raising an error will crash the front stats daemon and we don't want that. If something goes wrong we will still be able to see it via blank software versions component

kamtarnowski: Yeah, I haven't notice that this is running in front stats daemon.

kamtarnowski (approved):

kbiala (approved):

PR #784, opened 2023-04-12
184911359 - fix nodes counts

kamtarnowski:
#### What's this PR do?
PR fixes both total and delinquent values for stake pool nodes section.

- locally please  run `rails r script/gather_stake_accounts.rb` and check if no errors presented
- verify values in the "Stake Pools Overview" and each of the stake pool Statistics page, focussing on the Nodes section

[- URL to the PivotalTracker ticket](https://www.pivotaltracker.com/story/show/184911359)

#### Things to watch out for


app/models/stake_account.rb
```diff
@@ -68,6 +70,7 @@ class StakeAccount < ApplicationRecord
   scope :filter_by_staker, ->(staker) { where('staker LIKE ?', "#{staker}%") }
   scope :filter_by_withdrawer, ->(withdrawer) { where('withdrawer LIKE ?', "#{withdrawer}%") }
   scope :active, ->() { where.not(active_stake: [0, nil]) }
+  scope :greater_than_one_sol, ->() { where("active_stake > ?", MINIMUM_STAKE) }
```
maciejkocylapc, line 73: I would change this scope to take minimum stake as a parameter that defaults to MINIMUM_STAKE to be more versatile (and call it `having_minimum_stake` or something similar)

test/services/stake_pools/calculate_average_stats_test.rb
```diff
@@ -80,10 +80,10 @@ class CalculateAverageStatsTest < ActiveSupport::TestCase
 
       assert_equal 3, stake_pool.average_uptime
       assert_equal 10, stake_pool.average_lifetime
-      assert_equal 0, stake_pool.average_delinquent
       assert_equal 2.55, stake_pool.average_skipped_slots
       assert_equal 7, stake_pool.average_score
-      assert_equal 8.15, stake_pool.average_validators_commission
+      assert_equal 6.1, stake_pool.average_validators_commission
```
maciejkocylapc, line 85: why is this value changed?

maciejkocylapc (changes requested):

app/models/stake_account.rb
kamtarnowski: Sounds good. I did use the `with_minimum_stake`.

test/services/stake_pools/calculate_average_stats_test.rb
kamtarnowski: `active_stake` value has been changed and it has an impact on average_commission calculation.

maciejkocylapc (approved):

app/services/stake_pools/calculate_average_stats.rb
```diff
@@ -33,6 +33,10 @@ def call
 
     attr_reader :pool, :validator, :validator_ids, :sp_attrs
 
+    def delinquent?(score)
```
kbiala, line 36: Do not use method name that is the same as attribute name, but does a different thing. It's a bad practice. This method name should match its behaviour.

test/models/stake_pool_test.rb
```diff
@@ -9,29 +9,32 @@ class StakePoolTest < ActiveSupport::TestCase
...
+
+    create(:stake_account,
+           validator: @validator3,
+           stake_pool: @stake_pool,
+           active_stake: 0,
+           delegated_stake: 1_000_000_000_000)
   end
 
-  test "#validators_count returns total number of validators that pool delegates to" do
+  test "#validators_count returns total number of active validators that pool delegates to" do
```
kbiala, line 37: Not only active, but also with minimum stake, right? Update test description

test/services/stake_pools/calculate_average_stats_test.rb
```diff
@@ -67,7 +67,7 @@ class CalculateAverageStatsTest < ActiveSupport::TestCase
         stake_pool: stake_pool,
         delegated_vote_account_address: "vote_acc",
         validator: validator,
-        active_stake: 422_555_666
+        active_stake: 2_555_666_777
```
kbiala, line 70: What the reason for such a drastic value change?

kbiala (changes requested):

app/javascript/packs/stake_accounts/pool_stats.js
```diff
@@ -47,12 +47,20 @@ var StakePoolStats = Vue.component('StakePoolStats', {
               </span>
             </div>
             <div>
-              <span class="text-muted">Total:&nbsp;</span>
+              <span class="text-muted">
+                Total
+                <small>(> 1 SOL)</small>
+                :&nbsp;
+              </span>
```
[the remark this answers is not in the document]
kamtarnowski: Fixed.

app/models/stake_account.rb
```diff
@@ -37,6 +38,8 @@
 
 class StakeAccount < ApplicationRecord
 
+  MINIMUM_STAKE = 1_000_000_000.freeze # 1 sol
```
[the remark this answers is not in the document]
kamtarnowski: I forgot to remove it from `StakeAccountQuery` model.

app/models/stake_account.rb
```diff
@@ -68,6 +71,7 @@ class StakeAccount < ApplicationRecord
   scope :filter_by_staker, ->(staker) { where('staker LIKE ?', "#{staker}%") }
   scope :filter_by_withdrawer, ->(withdrawer) { where('withdrawer LIKE ?', "#{withdrawer}%") }
   scope :active, ->() { where.not(active_stake: [0, nil]) }
+  scope :with_minimum_stake, ->() { where("active_stake > ?", MINIMUM_STAKE) }
```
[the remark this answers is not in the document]
kamtarnowski: I used the `StakeAccount::MINIMUM_STAKE` constant.

test/services/stake_pools/calculate_average_stats_test.rb
kamtarnowski: This way the delinquent_count is covered in test, since stake_account has more than 1 SOL and score is delinquent. As result the validator is considered as a delinquent.

app/services/stake_pools/calculate_average_stats.rb
```diff
@@ -33,6 +36,10 @@ def call
 
     attr_reader :pool, :validator, :validator_ids, :sp_attrs
 
+    def validator_delinquent?(score)
```
kbiala, line 39: This new name is not much better. This method is NOT checking if validator is delinquent (delinquent attribute stores that information). This method is checking two conditions, just one of them happens to be related to deliquency.
Should be called something like `delinquent_with_minimum_stake` or similar.

kbiala (changes requested):
I get 500 on stake pools page with the following error:
```
Completed 500 Internal Server Error in 2ms (ActiveRecord: 0.4ms | Allocations: 3307)
ActiveRecord::StatementInvalid (Mysql2::Error: Column 'active_stake' in where clause is ambiguous):
app/controllers/api/v1/stake_accounts_controller.rb:57:in `group_by'
app/controllers/api/v1/stake_accounts_controller.rb:57:in `records_for_vue_frontend'
app/controllers/api/v1/stake_accounts_controller.rb:50:in `get_correct_records'
app/controllers/api/v1/stake_accounts_controller.rb:8:in `index'
```

kbiala (changes requested):
I don't get 500 now, but it still doesn't work correctly. See examples below.
For DAOpool, there are 9 validators >1, but none of them is delinquent, so delinqunt_count should be 0. There is 1 delinquent validator, but with stake below 1 SOL, so should not be included.
Zrzut ekranu 2023-04-25 o 17 09 08

Second case, socean pool. Again, none of the validators > 1 SOL is delinquent. Delinquent_count shows 1.
Zrzut ekranu 2023-04-25 o 17 07 23

kbiala (changes requested):
Also, please take a look at tests, and fix or add missing ones. These incorrect delinqent_counts should have been caught by tests.

kamtarnowski:
I confirm the issue regarding delinquent count. It was fixed by adding an extra scope for stake accounts.

PR #793, opened 2023-04-18
184961000 fix distance stats

maciejkocylapc:
#### What's this PR do?
- vote distance and root distance stats on home page should include only active validators

- test front stats update daemon if it returns no errors
- run tests
- test on staging if the values look correct

- [URL to the PivotalTracker ticket](https://www.pivotaltracker.com/story/show/184961000)

#### Things to watch out for

### DIFF
--- a/app/models/stats/validator_score.rb
+++ b/app/models/stats/validator_score.rb
@@ -56,7 +56,9 @@ module Stats
     end
 
     def vote_distance_all_history
-      @vote_distance_all_history ||= relation&.joins(:validator)&.pluck(:vote_distance_history, :account)
+      @vote_distance_all_history ||= relation&.joins(:validator)
+                                             &.where("validators.is_active = ? AND validator_score_v1s.delinquent = ?", true, false)
+                                             &.pluck(:vote_distance_history, :account)
     end
 
     def vote_distance_all_averages
--- a/app/models/stats/validator_score.rb
+++ b/app/models/stats/validator_score.rb
@@ -22,6 +22,7 @@ module Stats
 
     def root_distance_all_history
       @root_distance_all_history ||= relation.joins(:validator)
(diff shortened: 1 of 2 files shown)


app/models/stats/validator_score.rb
```diff
@@ -22,6 +22,7 @@ def initialize(network, batch_uuid)
 
     def root_distance_all_history
       @root_distance_all_history ||= relation.joins(:validator)
+                                             .where("validators.is_active = ? AND validator_score_v1s.delinquent = ?", true, false)
                                              .pluck(:root_distance_history, :account)
```
kamtarnowski, line 26:
```suggestion
      @root_distance_all_history ||=
        relation.joins(:validator)
                     .where("validators.is_active = ? AND validator_score_v1s.delinquent = ?", true, false)
                     .pluck(:root_distance_history, :account)
```

app/models/stats/validator_score.rb
```diff
@@ -55,7 +56,9 @@ def root_distance_stats(with_history: false)
     end
 
     def vote_distance_all_history
-      @vote_distance_all_history ||= relation&.joins(:validator)&.pluck(:vote_distance_history, :account)
+      @vote_distance_all_history ||= relation&.joins(:validator)
+                                             &.where("validators.is_active = ? AND validator_score_v1s.delinquent = ?", true, false)
+                                             &.pluck(:vote_distance_history, :account)
```
kamtarnowski, line 61: Does the rescue method &. is required? Or maybe we can get rid of that.

```suggestion
      @vote_distance_all_history ||= 
        relation.joins(:validator)
                     .where("validators.is_active = ? AND validator_score_v1s.delinquent = ?", true, false)
                     .pluck(:vote_distance_history, :account)
```

PR #795, opened 2023-04-18
184855662 - remove inactive filters

kamtarnowski:
#### What's this PR do?
Remove inactive filters and TOTAL stake reference over DC pages.

- verify filters in DC pages( index, show and ASN related versions)
- Total Stake was renamed into Stake. Please verify that.

[- URL to the PivotalTracker ticket](https://www.pivotaltracker.com/story/show/184855662)

#### Things to watch out for

### DIFF
--- a/app/views/asns/show.html.erb
+++ b/app/views/asns/show.html.erb
@@ -41,6 +41,13 @@
             <td><%= @population %></td>
           </tr>
           <tr>
+            <td><strong>Stake</strong></td>
+            <td>
+              <%= number_with_delimiter(lamports_to_sol(@asn_stake).round(0)) %>
+              <span class="text-muted">
+                (<%= number_to_percentage((@asn_stake/@total_stake.to_f) * 100.0, precision: 2) %>)
+              </span>
+            </td>
           </tr>
         </tbody>
       </table>
--- a/app/views/asns/show.html.erb
+++ b/app/views/asns/show.html.erb
@@ -41,13 +41,6 @@
             <td><%= @population %></td>
           </tr>
           <tr>
-            <td><strong>Total Stake</strong></td>
-            <td>
-              <%= number_with_delimiter(lamports_to_sol(@asn_stake).round(0)) %>
-              <span class="text-muted">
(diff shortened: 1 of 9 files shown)


app/views/asns/show.html.erb
```diff
@@ -41,13 +41,6 @@
             <td><%= @population %></td>
           </tr>
           <tr>
-            <td><strong>Total Stake</strong></td>
```
kbiala, line 44: We do not want to remove this field, just rename it to "Stake" (see the screenshots in PT)

test/models/validator_test.rb
```diff
@@ -183,11 +181,11 @@ class ValidatorTest < ActiveSupport::TestCase
     assert result.last.delinquent?
   end
 
-  test "#filtered_by inactive excludes correct validators from collection" do
-    result = Validator.filtered_by(["inactive"])
+  test "#filtered_by active excludes correct validators from collection" do
```
kbiala, line 184: Please fix this test name. Filtering by active (or anything else), doesn't exclude any validators. On the contrary, it includes all validators that meet condition (in this case - validators that are active), but they also may or may not meet other conditions (eg. might be also active and private) - it doesn't matter, as long as they meet selected filter. While I understand what test author wanted to say, this test description is confusing and incorrect.
It should be something like `#filtered_by active includes all active validators in collection`

test/models/validator_test.rb
```diff
@@ -198,10 +196,10 @@ class ValidatorTest < ActiveSupport::TestCase
   end
 
   test "#filtered_by multiple parameters excludes correct validators from collection" do
```
kbiala, line 198: Same here. Should be something like `#filtered_by multiple parameters includes all validators that meet any of selected criteria"`

kbiala (changes requested):

app/views/asns/show.html.erb
kamtarnowski: Fixed.

kbiala (approved):

PR #802, opened 2023-04-24
184549107 - tweak calculating epoch duration

kamtarnowski:
#### What's this PR do?
Move epoch duration value into db and its related logic into a separated service.

- `rails db:migration`
- run `ClusterStats::UpdateEpochDurationService.new(network).call` for each network you want to test. Check and confirm that epoch_duration in ClusterStat was populated with some value
- run `rails r script/gather_stake_accounts.rb` and check for any errors, check pools APY/ROI
- run gather scripts and confirm that there're no errors in sidekiq, and gross ROI is updated with some value

[- URL to the PivotalTracker ticket](https://www.pivotaltracker.com/story/show/184549107)

#### Things to watch out for


maciejkocylapc (changes requested):

app/services/cluster_stats/update_epoch_duration_service.rb
```diff
@@ -0,0 +1,34 @@
...
+    def cluster_stat
+      ClusterStat.find_or_create_by(network: network)
+    end
+
+    def epochs
+      @epochs ||= EpochWallClock.where(network: network).last(AVG_EPOCH_SIZE)
+    end
+
+    def epoch_duration
+      (epochs.last.created_at - epochs.first.created_at) / AVG_EPOCH_SIZE
```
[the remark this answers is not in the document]
kamtarnowski: Ok

app/views/public/faq.en.html.erb
```diff
@@ -341,8 +341,8 @@
   (stake_account_rewards + vote_account_rewards / total_active_stake) * number_of_epochs_per_year
 </pre>
 <p>
-  Where number_of_epochs_per_year is calculated as follows: seconds_in_year / duration_of_last_epoch_in_seconds. 
-  This gives the number of epochs in a year assuming all epochs were as long as the last epoch.
+  Where number_of_epochs_per_year is calculated as follows: seconds_in_year / average_duration_from_the_last_five_epochs_in_seconds.
```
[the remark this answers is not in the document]
kamtarnowski: Good catch. Changing it.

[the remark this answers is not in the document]
kbiala: Let's just call it average_epoch_duration in formula.
And add below `average_epoch_duration is based on ` and AVG_EPOCH_SIZE constant.

app/services/cluster_stats/update_epoch_duration_service.rb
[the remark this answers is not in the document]
kamtarnowski: I reverted changes, since `datetime - datetime` returns a float. `to_f` is redundant in this case.

app/services/cluster_stats/update_epoch_duration_service.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+module ClusterStats
+  class UpdateEpochDurationService
+
+    AVG_EPOCH_SIZE = 5
```
kbiala, line 6: What does this constant represent? It's not epoch size, all the more average epoch size. It's a number of epochs that we pull for calculations, so should be called appropriately, NUMBER_OF_EPOCHS.

app/services/cluster_stats/update_epoch_duration_service.rb
```diff
@@ -0,0 +1,34 @@
...
+    private
+
+    attr_reader :network
+
+    def cluster_stat
+      ClusterStat.find_or_create_by(network: network)
+    end
+
+    def epochs
+      @epochs ||= EpochWallClock.by_network(network).last(AVG_EPOCH_SIZE)
```
kbiala, line 27: by_network scope sorts epochs in descending order, so you should take first 5 records, not last. Please check how it used in other places of the app next time.

app/logic/stake_logic.rb
```diff
@@ -217,7 +217,7 @@ def calculate_apy_for_accounts
       return p unless p.code == 200
 
       # number of epochs in year calculated using the duration of last finished epoch
-      num_of_epochs = number_of_epochs(p.payload[:current_epoch].created_at, p.payload[:previous_epoch].created_at)
+      num_of_epochs = number_of_epochs(p.payload[:network])
```
kbiala, line 220: You removed using current_epoch and previous_epoch, but left all the related code. Variables, set_epochs method, assign_epochs method. There's still a lot to do in this story.

app/services/create_cluster_stats_service.rb
```diff
@@ -36,17 +34,27 @@ def call
...
   def epochs_annual_roi
     if last_epochs.count > 0
-      epoch_duration = (last_epochs.first.created_at - last_epochs.second.created_at)
-      return 0 unless epoch_duration.to_f > 0
-      epoch_average_roi = last_epochs.map do |epoch|
-        if epoch.total_rewards && epoch.total_active_stake
-          (epoch.total_rewards / epoch.total_active_stake.to_f) * 100
-        end
-      end.compact&.average.to_f
+      epoch_average_roi =
```
kbiala, line 48: Never break the lines this way. If you want to pull the variable name before block, it should be aligned as below.
```suggestion
      epoch_average_roi = last_epochs.map do |epoch|      
```

app/services/create_cluster_stats_service.rb
```diff
@@ -36,17 +34,27 @@ def call
...
+        last_epochs.map do |epoch|
+          if epoch.total_rewards && epoch.total_active_stake
+            (epoch.total_rewards / epoch.total_active_stake.to_f) * 100
+          end
+        end.compact&.average.to_f
       return 0 unless epoch_average_roi > 0
-      epochs_annual_roi = (1.year.to_i / epoch_duration.to_f) * epoch_average_roi
+
+      epochs_annual_roi =
+        (1.year.to_i / epoch_duration) * epoch_average_roi
```
kbiala, line 57: Same here. It should be one line.

app/services/create_cluster_stats_service.rb
```diff
@@ -36,17 +34,27 @@ def call
 
   private
 
+  def network_stat
+    @network_stat ||= ClusterStat.find_or_create_by(network: @network)
+  end
```
kbiala, line 39: I can see a lot of changed not related to this story. Please avoid it in the future, especially if the story is big enough without additional refactoring. It makes it harder to find where you made a mistake.

kbiala (changes requested):

test/services/cluster_stats/update_epoch_duration_service_test.rb
```diff
@@ -0,0 +1,33 @@
...
+  class UpdateEpochDurationServiceTest < ActiveSupport::TestCase
+    setup do
+      @network = "testnet"
+
+      create(:epoch_wall_clock, epoch: 301, created_at: 16.days.ago)
+      create(:epoch_wall_clock, epoch: 311, created_at: 13.days.ago)
+      create(:epoch_wall_clock, epoch: 322, created_at: 10.days.ago)
+      create(:epoch_wall_clock, epoch: 333, created_at: 7.days.ago)
+      create(:epoch_wall_clock, epoch: 344, created_at: 4.days.ago)
+      create(:epoch_wall_clock, epoch: 355, created_at: 1.day.ago)
```
kbiala, line 15: Technically it's not an error, but why not use epochs numbers that actually represent epochs that happen one after another in time (whether 1, 2, 3, 4, 5 or 351, 352, 353, 354, 355, no difference). This is how it works in reality. Skipping numbers might be confusing for someone working on it in the future, for example might suggest that we want to skip some epochs, which it's not true.

kbiala (changes requested):
Tests are failing.

kamtarnowski:
I have addressed all of the raised issues.

test/services/cluster_stats/update_epoch_duration_service_test.rb
```diff
@@ -0,0 +1,34 @@
...
+      create(:epoch_wall_clock, epoch: 306, created_at: 1.day.ago)
+    end
+
+    test "updates epoch duration" do
+      cluster_stat = ClusterStat.last
+      cluster_stat.update(epoch_duration: 0)
+
+      ClusterStats::UpdateEpochDurationService.new(@network).call
+
+      assert_equal 207360.0, cluster_stat.reload.epoch_duration
```
kbiala, line 24: This is incorrect. Do not write tests that always pass, write tests to detect errors.
You created 6 epochs, every one lasting for about 3 days, and yet average is about 2.4 days. It's not correct. The calculated average should be much closer to ~260000 (3 days), and this is the result when I calculate it manually. There's an error somewhere in your code.

kbiala (changes requested):
Testing on staging showed that epoch average duration is not calculated correctly.
It calculated 163912.0 (1.89 days) for mainnet, 155846.0 (1.8 days) for testnet, and 133338.0 (1.54 days) for pythnet. The gross ROI is also messed up because of incorrect epoch duration.
Last 5 epoch on mainnet (starting from the newest one) lasted for 
2.38 days, 2.38 days, 2.39 days, 2.23 days and 2.31 days. So the average cannot be shorter than 2 days.

test/services/cluster_stats/update_epoch_duration_service_test.rb
kbiala: And you should also check all other places where you changes some values in tests. Looks like you made a lot of unjustified changes, just to make tests pass.

app/services/cluster_stats/update_epoch_duration_service.rb
```diff
@@ -0,0 +1,34 @@
...
+    def cluster_stat
+      ClusterStat.find_or_create_by(network: network)
+    end
+
+    def epochs
+      @epochs ||= EpochWallClock.by_network(network).first(NUMBER_OF_EPOCHS)
+    end
+
+    def epoch_duration
+      (epochs.first.created_at - epochs.last.created_at) / NUMBER_OF_EPOCHS
```
kbiala, line 31: Here's an error. If you take 5 epochs, you get 4 time differences.
You should take first 6 epochs in line 27, to divide by 5.

test/logic/stake_logic/apy_helper_test.rb
```diff
@@ -9,9 +9,11 @@ class ApyHelperTest < ActiveSupport::TestCase
   end
 
   test "number_of_epochs returns correct number" do
-    num_of_epochs = number_of_epochs(1.day.ago, 2.days.ago)
+    create(:cluster_stat, network: @network, epoch_duration: 159521.0)
```
kbiala, line 12: Please set a realistic epoch duration, so we can actually see if the results make sense. Epochs are never shorter than 2 days.

kbiala (changes requested):

kbiala:
There's still an old descript in Stake Pools docs

Zrzut ekranu 2023-05-11 o 17 05 57

kbiala (changes requested):

kamtarnowski:
Testing factories were changed to more realistic (epoch_duration).

The number of epochs, which we are using to calculate epoch duration, has increased to 6, so that there we can utilize 5 values of time differences between epochs.

test/services/cluster_stats/update_epoch_duration_service_test.rb
```diff
@@ -0,0 +1,49 @@
...
+      create(:epoch_wall_clock, epoch: 307)
+    end
+
+    test "updates epoch duration" do
+      cluster_stat = ClusterStat.last
+      cluster_stat.update(epoch_duration: 0)
+
+      ClusterStats::UpdateEpochDurationService.new(@network).call
+
+      assert_equal 180000.0, cluster_stat.reload.epoch_duration
```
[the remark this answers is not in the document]
kamtarnowski: I fixed that and the value is 216000.0 currently.

test/services/cluster_stats/update_epoch_duration_service_test.rb
```diff
@@ -0,0 +1,49 @@
...
+    setup do
+      @network = "testnet"
+
+      create(:epoch_wall_clock, epoch: 301, created_at: 16.days.ago)
+      create(:epoch_wall_clock, epoch: 302, created_at: 12.5.days.ago)
+      create(:epoch_wall_clock, epoch: 303, created_at: 10.days.ago)
+      create(:epoch_wall_clock, epoch: 304, created_at: 7.5.days.ago)
+      create(:epoch_wall_clock, epoch: 305, created_at: 5.days.ago)
+      create(:epoch_wall_clock, epoch: 306, created_at: 2.5.days.ago)
+      create(:epoch_wall_clock, epoch: 307)
```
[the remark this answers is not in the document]
kamtarnowski: I added one more epoch_wall_clock to check whether the calculation takes in proper number of epochs( 6). I will remove it.

app/services/cluster_stats/update_epoch_duration_service.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    def epochs
+      @epochs ||= EpochWallClock.by_network(network).first(NUMBER_OF_EPOCHS)
+    end
+
+    def epoch_duration
+      diff = epochs.first.created_at - epochs.last.created_at
+      return diff if epochs.size <= 2
+
+      diff / epochs.size
```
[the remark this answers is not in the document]
kamtarnowski: I thought about that solution at some point, but somehow I didn't apply that...

app/services/cluster_stats/update_epoch_duration_service.rb
```diff
@@ -0,0 +1,37 @@
...
+      ClusterStat.find_or_create_by(network: network)
+    end
+
+    def epochs
+      @epochs ||= EpochWallClock.by_network(network).first(NUMBER_OF_EPOCHS)
+    end
+
+    def epoch_duration
+      diff = epochs.first.created_at - epochs.last.created_at
+      return diff if epochs.size <= 2
```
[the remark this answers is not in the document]
kamtarnowski: I did apply your solution.

app/services/cluster_stats/update_epoch_duration_service.rb
```diff
@@ -0,0 +1,38 @@
...
+      ClusterStat.find_or_create_by(network: network)
+    end
+
+    def epochs
+      @epochs ||= EpochWallClock.by_network(network).first(NUMBER_OF_EPOCHS)
+    end
+
+    def epoch_duration
+      diff = epochs.first.created_at - epochs.last.created_at
+      diffs_between_epochs_amount = epochs.size - 1
```
kbiala, line 32: If you must create a separate variable, which isn't really needed in this case, at least call it appropriately, eg. diff_between_epochs_sum
And then give diff variable (line 31) more meaning name as well, eg last_epochs_duration

kbiala (changes requested):

kbiala (approved):

PR #805, opened 2023-04-25
184977223 avoid duplicating dc keys

maciejkocylapc:
#### What's this PR do?
- merge data centers with duplicated keys by moving all validators and gossip nodes to a single data center
- avoid creating duplicates by reducing fields by which existing datacenter is searched for

- test on staging

- [URL to the PivotalTracker ticket](https://www.pivotaltracker.com/story/show/184977223)

#### Things to watch out for


app/services/data_centers/merge_duplicates.rb
```diff
@@ -2,203 +2,59 @@ module DataCenters
...
-          log_processed_data_center(entry)
-
-          entry.data_center.data_center_hosts.each do |dch|
-            log_processed_data_center_host(dch)
-
-            main_dc_host = find_or_create_host(main_dc, dch)
+      duplicated_data_center_keys.each do |dc_key|
+        puts dc_key
+        sorted_data_centers = sort_data_centers_by_created_at(dc_key)
+        main_dc = sorted_data_centers.last
```
kbiala, line 16: I'd put main_dc here, so that user can check if correct DC was detected.
Or even better, all sorted_data_centers dates.

app/services/data_centers/merge_duplicates.rb
```diff
@@ -2,203 +2,59 @@ module DataCenters
...
           end
         end
-
-        log_message("---------------", type: :info)
+        switch_validator_ips(main_dc_host)
+        puts "switched validators and nodes: #{@validator_ips_to_switch.count}"
       end
-      log_message("---------------SCRIPT HAS FINISHED----------------", type: :info)
+      detach_validator_ips
+      puts "detached validator_ips: #{@validator_ips_to_detach.count}"
```
kbiala, line 34: Same here.

app/services/data_centers/merge_duplicates.rb
```diff
@@ -2,203 +2,59 @@ module DataCenters
...
+        puts dc_key
+        sorted_data_centers = sort_data_centers_by_created_at(dc_key)
+        main_dc = sorted_data_centers.last
+        main_dc_host = main_dc.data_center_hosts.last
+        @validator_ips_to_switch = []
+
+        sorted_data_centers.each do |dc|
+          next if dc == main_dc
+          dc.validators.joins(:validator_ips).each do |val|
+            append_ip_to_update(val.validator_ip_active, val.is_active)
```
kbiala, line 23: val.validator_ip_active? 
We want to reattach all IPs, not only active (not used anymore, that's why we want to switch them manually). Otherwise, we won't be able to remove old DataCenters, there still will be old IPs attached to them.

app/services/data_centers/merge_duplicates.rb
```diff
@@ -2,203 +2,59 @@ module DataCenters
...
+          end
 
-            dch.validator_ips.each do |vip|
-              update_validator_ip(vip, main_dc_host)
-            end
+          dc.gossip_nodes.joins(:validator_ip).each do |node|
+            append_ip_to_update(node.validator_ip_active, node.is_active)
```
kbiala, line 27: Same for gossip nodes.

kbiala (changes requested):

app/services/data_centers/merge_duplicates.rb
kbiala: Actually, now I see that you're trying to pull the list of IPs via validators in line 22, which is not correct. What you should do instead, you should pull the list of IPs attached to data center, so `dc.validator_ips`, which will give you all IPs (and dc.validator_ips_active will give you only active ones, but we want all).

kamtarnowski (commented):
@kbiala caught appropriate issues.

Let's try to make more appropriate argument names, as well as adding a version script, which will only show the further db changes (run_update)

kbiala (approved):

PR #807, opened 2023-05-02
184866369 cleanup validator check active service

maciejkocylapc:
#### What's this PR do?
- refactor ValidatorCheckActiveService
- keep similar functionality

- run tests
- try to update validators is_active, is_destroyed and is_rpc and see if the script updates it correctly

- [URL to the PivotalTracker ticket](https://www.pivotaltracker.com/story/show/184866369)

#### Things to watch out for

### DIFF
--- a/app/services/validator_check_active_service.rb
+++ b/app/services/validator_check_active_service.rb
@@ -13,25 +13,26 @@ class ValidatorCheckActiveService
       # Skip if validator has no history that is older than DELINQUENT_TIME
       next if too_young?(validator)
 
+      # mark validator as destroyed if it has no recent history
       if should_be_destroyed?(validator)
-        validator.update(is_active: false, is_destroyed: true)
-      elsif validator.scorable?
-        update_scorable(validator)
+        validator.assign_attributes(is_active: false, is_destroyed: true)
       else
-        # Check validators that are currently not scorable in case they reactivate
-        if acceptable_stake?(validator) && !delinquent?(validator)
-          validator.update(is_active: true, is_destroyed: false)
-        end
-        
-        # Check if vote account has been created for rpc_node
-        if validator.vote_accounts.exists?
-          validator.update(is_rpc: false) if validator.is_rpc
-        end
+        validator.assign_attributes(is_destroyed: false)
+        update_scorable(validator)
       end
+
+      validator.save if validator.changed?
     end
   end
 
   private
+
+  # returns true if validator has no history from previous epoch
+  def too_young?(validator)
+    !validator.validator_histories
+              .where("created_at < ?", @delinquent_time.ago)
+              .exists?
+  end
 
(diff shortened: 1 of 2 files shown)


app/services/validator_check_active_service.rb
```diff
@@ -99,6 +78,7 @@ def acceptable_stake?(validator)
     with_acceptable_stake.exists?
   end
 
+  # check if validator had no delinquent history since DELINQUENT_TIME
   def non_delinquent_history_exists?(validator)
     ValidatorHistory.where(account: validator.account, delinquent: false)
```
kbiala, line 83: Can we use `validator_histories_for_validator` method here as well? Or edit validator_histories_for_validator method so that it can be reused here?

app/services/validator_check_active_service.rb
```diff
@@ -10,42 +10,22 @@ def initialize
 
   def update_validator_activity
     Validator.includes(:vote_accounts, :validator_score_v1).find_each do |validator|
+      # Skip if validator has no history that is older than DELINQUENT_TIME
+      next if too_young?(validator)
+
+      # mark validator as destroyed if it has no recent history
       if should_be_destroyed?(validator)
```
kbiala, line 17: I don't really see a point of keeping this method, since it's the same as simply writing `if validator_histories_for_validator(validator).empty?`, and it's only used once.

app/services/validator_check_active_service.rb
```diff
@@ -54,34 +34,33 @@ def too_young?(validator)
...
-  def update_scorable(validator)
-    if !too_young?(validator) && became_inactive?(validator)
-      validator.update(is_active: false)
-    end
+  def should_be_rpc?(validator)
+    !validator.vote_accounts.exists?
+  end
 
-    validator.update(is_rpc: true) if is_rpc?(validator)
+  def update_scorable(validator)
```
kbiala, line 56: Why is this method called update_scorable?

app/services/validator_check_active_service.rb
```diff
@@ -54,34 +34,33 @@ def too_young?(validator)
...
+      is_active: should_be_active?(validator),
+      is_rpc: should_be_rpc?(validator)
+    )
   end
 
-  def delinquent?(validator)
+  def long_time_delinquent?(validator)
     return false if !validator.delinquent?
 
     !non_delinquent_history_exists?(validator)
```
kbiala, line 66: Same here, I'd use validator_histories_for_validator method explicitly here, instead of extracting it to another method, because both `non_delinquent_history_exists` and `should_be_destroyed` methods don't add any extra logic at all, they just call another method, and nothing else. So there's no reason for them to exist.

kbiala (changes requested):

kbiala (approved):

PR #810, opened 2023-05-11
184205895 api for authorities changes

maciejkocylapc:
#### What's this PR do?
- api for account authority histories

- run tests
- run requests against `/api/v1/account-authorities/:network` and make sure everything works properly
- check API docs

- [URL to the PivotalTracker ticket](https://www.pivotaltracker.com/story/show/184205895)

#### Things to watch out for


app/views/public/api_documentation.html.erb
```diff
@@ -600,6 +601,45 @@
...
+<p><strong class="text-muted">`per=NN`</strong> will limit the results to NN entries per page, defaults to 100.</p>
+<pre>curl -H "Token: secret-api-token" '<%= @api_path %>/account-authorities/:network.json?per=50'</pre>
+<p><strong class="text-muted">`page=NN`</strong> will return the results from provided page, defaults to 1.</p>
+<pre>curl -H "Token: secret-api-token" '<%= @api_path %>/account-authorities/:network.json?per=50&page=1'</pre>
+
+<h3 class="h6">Response:</h3>
+<pre>
+[
+	{
+		"id": 175,
```
kbiala, line 624: We could omit our internal ID in response

app/views/public/api_documentation.html.erb
```diff
@@ -600,6 +601,45 @@
...
+		"authorized_withdrawer_before": null,
+		"authorized_withdrawer_after": "4hEF3wqsyr1NxXf2SYec3Qs8yk24vWsT7o7Cw7HsTLej",
+		"authorized_voters_before": null,
+		"authorized_voters_after": {
+			"432": "DDnAqxJVFo2GVTujibHt5cjevHMSE9bo8HJaydHoshdp"
+		},
+		"vote_account_id": 176,
```
kbiala, line 631: We should return vote account identity instead of vote_account_id in response, because our internal ID woun't tell people much. They won't know which VA this record is related to.

app/models/account_authority_history.rb
```diff
@@ -27,4 +27,14 @@ class AccountAuthorityHistory < ApplicationRecord
...
   serialize :authorized_voters_after, JSON
+
+  API_FIELDS = %i[
+    authorized_voters_after
+    authorized_voters_before
+    authorized_withdrawer_after
+    authorized_withdrawer_before
+    network
+    created_at
+    updated_at
```
kbiala, line 38: In this case I think created_at would be enough. But OK to leave it if you think it might be useful in some cases.

kbiala (changes requested):
The code looks good, just a few small comments.
And one more thing after QA:
I get results (last 100 records by default), even if I pass incorrect validator/vote account param, or when I pass validator as VA, validator from different network etc. I think we should return empty response in that case (becaues no records match given criteria).

kbiala:
Looks good, except I still got results if I pass incorrect criteria (eg. non existing validator identity). Response should be empty in this case.

PR #817, opened 2023-05-19
184205913 - add authorities to vote account view

kamtarnowski:
#### What's this PR do?
PR adds the authorized withdrawers and voters to the vote account details page.

- evaluate the design of new components and its content
- verify it also on various resolutions (web, mobile)
- check if the new content appears when there aren't any records
- both authorized withdrawers and voters should be displayed in descending order as well as they should be unique.

[- URL to the PivotalTracker ticket](https://www.pivotaltracker.com/story/show/184205913)

#### Things to watch out for


kbiala (changes requested):
The last commit completely broke this feature. See screenshots below.
We DO NOT want to show unique records, but all records in history. If the validator used authority A, then changed to B, then back to A (it happens often), all 3 changes should be shown. Instead, you show one.
Zrzut ekranu 2023-05-24 o 16 44 15

Zrzut ekranu 2023-05-24 o 16 44 06

kamtarnowski:
> The last commit completely broke this feature. See screenshots below. We DO NOT want to show unique records, but all records in history. If the validator used authority A, then changed to B, then back to A (it happens often), all 3 changes should be shown. Instead, you show A only once. It's incorrect You can compare before and after values to decide if the row should be displayed, but not the way you did here.
> 
> Zrzut ekranu 2023-05-24 o 16 44 15 Zrzut ekranu 2023-05-24 o 16 44 06

kamtarnowski:
https://github.com/brianlong/www.validators.app/pull/817#pullrequestreview-1442092465

Currently, when saving any changes to either voters or withdrawer will cause saving both attributes. So there is probably situation when in db we have:

Account authority history record with following attrs
authorized_voters: value_1
authorized_withdrawer: value_2

After vote account update the Account authority history record is created
authorized_voters: value_1
authorized_withdrawer: value_3

As result there wasn't any changes in terms of authorized_voters, so it isn't important information for the user. Am I right? The duplicated records should be skipped?

kbiala:
Just compare before and after value of the same record. And you can skip displaying if it's the same (and only then). But honestly, it wouldn't hurt to just display them (assuming all is in the same table).
But that's not the case that I described earlier. You removed all non unique records from the history, and it's completely not what we want.

kbiala (changes requested):

kbiala (changes requested):
After downloading recent code, I can no longer see table on any vote account page.
All pages return an error
Zrzut ekranu 2023-05-30 o 13 29 35

PR #825, opened 2023-06-06
185341569 - add validator_id to validator history

kamtarnowski:
#### What's this PR do?
Add a validator association to validator histories. For minitest tests the only change is that the validator_id were injected, except one added test covering a new script.

- `rails test`
- check if none of the ValidatorHistory records locally does not have validator_id assigned
- execute and compare results for `rails r script/one_time_scripts/add_validator_id_to_validator_histories.rb`. It shouldn't rise any errors. All validator histories should now have validator_id assigned in, except those one that the validator does not exist anymore
- I had about 200k records and it took about ~1 minute to process in.

[- URL to the PivotalTracker ticket](https://www.pivotaltracker.com/story/show/185341569)

#### Things to watch out for


app/logic/solana_logic.rb
```diff
@@ -651,9 +653,15 @@ def check_epoch
 
   def existing_validators(validator_attrs, network)
     accounts = validator_attrs.map { |validator| validator["identityPubkey"] }
-    db_validator_accounts = Validator.where(network: network, account: accounts).pluck(:account)
-    validator_attrs.select do |validator|
-      db_validator_accounts.include?(validator["identityPubkey"])
-    end
+    db_validator_accounts = Validator.where(network: network, account: accounts)
```
maciejkocylapc, line 656: I don't think those changes are necessary - there's no reason why we should use ids here

maciejkocylapc (changes requested):

app/logic/solana_logic.rb
kamtarnowski: We need to extract those ids in order to add validator relation to validator history in `validator_history_update` method.

maciejkocylapc (approved):

kamtarnowski:
WIP with `disable_ddl_transaction`.

PR #826, opened 2023-06-06
185302616 - fix saving Authorized Voters change

kamtarnowski:
#### What's this PR do?
PR applies changes to callback logic for authorized_voters. New AccountAuthorityHistory record should be created if authorized_voters hash values have changed. New one time script will remove invalid authorized voters histories.

- execute `rails r script/one_time_scripts/remove_invalid_authorized_voters.rb` and look for any errors
- vote accounts should have most recent authorized_voters value (from latest history)
- verify via console if there are repeated values of authorized_voters hash within each VoteAccount record. If so then the script has failed.

[- URL to the PivotalTracker ticket](https://www.pivotaltracker.com/story/show/185302616)

#### Things to watch out for

### DIFF
--- a/script/one_time_scripts/remove_invalid_authorized_voters.rb
+++ b/script/one_time_scripts/remove_invalid_authorized_voters.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+require File.expand_path('../../config/environment', __dir__)
+
+VoteAccount.joins(:account_authority_histories)
+           .includes(:account_authority_histories)
+           .find_each do |va|
+  invalid_voters = va.account_authority_histories.select do |history|
+    if history.authorized_withdrawer_after == history.authorized_withdrawer_before
+      before_values = Array(history.authorized_voters_before&.values)
+      after_values = Array(history.authorized_voters_after&.values)
+
+      before_values.size == (before_values & after_values).size
+    end
+  end
+
+  AccountAuthorityHistory.where(id: invalid_voters).destroy_all
+
+  history_last = va.account_authority_histories.last
+
+  if va.authorized_voters != history_last.authorized_voters_after ||
+    va.authorized_withdrawer != history_last.authorized_withdrawer_after
+    va.update_columns(
+      authorized_voters: history_last.authorized_voters_after,
+      authorized_withdrawer: history_last.authorized_withdrawer_after
+    )
+  end
+end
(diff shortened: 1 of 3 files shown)


script/one_time_scripts/remove_invalid_authorized_voters.rb
```diff
@@ -0,0 +1,24 @@
...
+
+VoteAccount.joins(:account_authority_histories)
+           .includes(:account_authority_histories)
+           .find_each do |va|
+  invalid_voters = va.account_authority_histories.select do |history|
+    if history.authorized_withdrawer_after == history.authorized_withdrawer_before
+      before_values = Array(history.authorized_voters_before&.values)
+      after_values = Array(history.authorized_voters_after&.values)
+
+      before_values.size == (before_values & after_values).size
```
kbiala, line 13: This is incorrect.
You're comparing only to before_values, so It will only work in the scenario:
before_values = ["test"]
after_values = ["test"] or after_values = ["test2"]  (equal number of elements before and after)

or
before_values = ["test", "test2"]
after_values = ["test"]
(more elements before than after)

but will not work in the following scenario:
before_values = ["test"]
after_values = ["test", "test2"]
(more elements after than before)
because the common part of the collection ("test") is the same as before_values. And before and after values are not at all the same, so it incorrectly detects this record to be removed. 

Instead, you should just sort and compare arrays.

Zrzut ekranu 2023-06-13 o 21 46 08


And please, write some tests for your code. This kind of issues should be detected by tests.

kbiala (changes requested):

script/one_time_scripts/remove_invalid_authorized_voters.rb
kamtarnowski: Good catch. Agree, sorting and comparing way will do the thing, but I see we can remove one array from the another and see if any elements has left. This seems slightly faster.

kamtarnowski: Good catch. Agree, sorting and comparing way will do the thing, but I see we can use `<=>` operator as well in order to compare arrays, which seems slightly faster.

kbiala: Do not substract arrays (I'm referring to your previous comment) - this solutions is prone to errrors, because it matters  which array is larger.
Comparing with spaceship operator will work, but remember it returns -1/0/1, instead of true/false. Also, you won't avoid sorting.

kamtarnowski: I cannot avoid sorting. That's a pitty, but it's true.

kbiala (approved):

PR #829, opened 2023-06-21
185395899 mark jito validators

maciejkocylapc:
#### What's this PR do?
- add jito mark to validators in validators table and show jito commission in detail view

- run migration
- run script/update_jito_validators.rb
- check validators table and details

#### Things to watch out for

### DIFF
--- a/script/update_jito_validators.rb
+++ b/script/update_jito_validators.rb
@@ -9,8 +9,8 @@ response = Net::HTTP.get_response(uri)
 if response.is_a?(Net::HTTPSuccess)
   json_response = JSON.parse(response.body)
   jito_vote_accounts = json_response["validators"].map do |jito_validator|
-    [jito_validator["vote_account"], jito_validator["mev_commission_bps"]] if jito_validator["running_jito"]
-  end
+    { jito_validator["vote_account"] => jito_validator["mev_commission_bps"] } if jito_validator["running_jito"]
+  end.inject(:merge)
 
   jito_db_validators = Validator.joins(:vote_accounts)
                                 .where(
@@ -18,14 +18,12 @@ if response.is_a?(Net::HTTPSuccess)
                                    AND vote_accounts.account IN (?)
                                    AND vote_accounts.is_active = TRUE",
                                   "mainnet",
-                                  jito_vote_accounts.map { |jito_vote_account| jito_vote_account[0] }
+                                  jito_vote_accounts.keys
                                 )
 
   jito_db_validators.each do |validator|
     validator.update(
-      jito_commission: jito_vote_accounts.find { |jito_vote_account| 
-        jito_vote_account[0] == validator.vote_account_active.account 
-      }[1],
+      jito_commission: jito_vote_accounts[validator.vote_account_active.account],
       jito: true
     )
   end
--- a/script/update_jito_validators.rb
+++ b/script/update_jito_validators.rb
@@ -24,7 +24,7 @@ if response.is_a?(Net::HTTPSuccess)
   jito_db_validators.each do |validator|
     validator.update(
       jito_commission: jito_vote_accounts.find { |jito_vote_account| 
-        jito_vote_account[0] == validator.vote_accounts.first.account 
+        jito_vote_account[0] == validator.vote_account_active.account 
       }[1],
       jito: true
     )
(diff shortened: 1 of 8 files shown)


script/update_jito_validators.rb
```diff
@@ -18,9 +18,17 @@
...
                                   "mainnet",
-                                  jito_vote_accounts
+                                  jito_vote_accounts.map { |jito_vote_account| jito_vote_account[0] }
                                 )
 
-  jito_db_validators.update_all(jito: true)
+  jito_db_validators.each do |validator|
+    validator.update(
+      jito_commission: jito_vote_accounts.find { |jito_vote_account| 
+        jito_vote_account[0] == validator.vote_accounts.first.account
```
kbiala, line 27: Shouldn't that be `validator.vote_account_active` ?

kbiala (changes requested):
I get the following error when I try to run on localhost:
```
script/update_jito_validators.rb:27:in `block in <main>': undefined method `[]' for nil:NilClass (NoMethodError)
```
I suspect that the cause of this is that you're searching by active vote accounts in the first query, but then selecting the first validator's vote account in the second query, instead of the active one.

script/update_jito_validators.rb
maciejkocylapc: fixed and added some test

test/scripts/update_jito_validators_test.rb
```diff
@@ -6,6 +6,9 @@ class UpdateJitoValidatorsTest < ActiveSupport::TestCase
   setup do
     @network = "mainnet"
     @validator = create(:validator, network: @network)
+    create(:vote_account,
+      account: "E3yhPs5PPN4RZh8FbJo2eqtdrAYKCK9H7pcSinactive",
+      validator: @validator, is_active: false)
```
kamtarnowski, line 11:
```
    create(:vote_account,
           account: "E3yhPs5PPN4RZh8FbJo2eqtdrAYKCK9H7pcSinactive",
           validator: @validator,
           is_active: false)
```
or
```
    create(
      :vote_account,
      account: "E3yhPs5PPN4RZh8FbJo2eqtdrAYKCK9H7pcSinactive",
      validator: @validator,
      is_active: false
)
  ```

File: script/update_jito_validators.rb [hunk 3]
```diff
@@ -9,7 +9,7 @@
 if response.is_a?(Net::HTTPSuccess)
   json_response = JSON.parse(response.body)
   jito_vote_accounts = json_response["validators"].map do |jito_validator|
-    jito_validator["vote_account"] if jito_validator["running_jito"]
+    [jito_validator["vote_account"], jito_validator["mev_commission_bps"]] if jito_validator["running_jito"]
   end
```
kamtarnowski, line 13: Let's use a Hash instead of an Array. That would remove some further logic below.

 ```
  jito_vote_accounts = json_response["validators"].map do |validator|
{ validator["vote_account"] => validator["mev_commission_bps"] } if validator["running_jito"]
end

```

File: script/update_jito_validators.rb [hunk 4]
```diff
@@ -18,9 +18,17 @@
                                    AND vote_accounts.account IN (?)
                                    AND vote_accounts.is_active = TRUE",
                                   "mainnet",
-                                  jito_vote_accounts
+                                  jito_vote_accounts.map { |jito_vote_account| jito_vote_account[0] }
```
kamtarnowski, line 21: And then just put on here:
`jito_vote_accounts.keys`

script/update_jito_validators.rb
```diff
@@ -18,9 +18,17 @@
...
                                 )
 
-  jito_db_validators.update_all(jito: true)
+  jito_db_validators.each do |validator|
+    validator.update(
+      jito_commission: jito_vote_accounts.find { |jito_vote_account| 
+        jito_vote_account[0] == validator.vote_account_active.account 
+      }[1],
```
kamtarnowski, line 28: After applying a Hash please do also:
```suggestion
      jito_commission: jito_vote_accounts[validator.vote_account_active.account],
```

kamtarnowski (changes requested):

kamtarnowski:
Css for the Jito icons in index and details page are not the same:

Validators#index:
Zrzut ekranu 2023-06-26 o 10 14 32

Validators#details_page
Zrzut ekranu 2023-06-26 o 10 14 38

kamtarnowski:
What about the case when jito_commission has value 0? Does it still should be marked as Jito and have an icon?

kamtarnowski (approved):

PR #857, opened 2023-09-19
183532936 gather all stake accounts

maciejkocylapc:
#### What's this PR do?
- script for gathering stake accounts for explorer

- run tests
- make sure You have any EpochWallClocks
- run `rails r script/gather_stake_accounts.rb`
- alternatively you can run the service in ruby console `GatherExplorerStakeAccountsService.new(network: "mainnet", config_urls:, current_epoch: nil, demo: true, stake_accounts: []).call`

- [URL to the PivotalTracker ticket](https://www.pivotaltracker.com/story/show/183532936)

#### Things to watch out for


kbiala (changes requested):

script/gather_stake_accounts.rb
```diff
@@ -24,6 +24,17 @@
 
   TotalRewardsUpdateService.new(p.payload[:network], p.payload[:stake_accounts_active]).call
 
+  if (StakeAccount.where(network: network).last&.epoch || 0) > \
```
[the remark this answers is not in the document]
kbiala: Also, what happens if last StakeAccount has not been updated - eg. because it is no longer active and no longer returned by Solana CLI? StakeAccount.last.epoch will then have an old epoch number, not current epoch. I have 15 records like this on my local machine (not the last one, but it could theoretically be the last one?). It will probably never happen in reality, but I think would be better to pull current epoch from EpochWallClock (EpochWallClock.by_network...first.epoch) instead of StakeAccount? And then compare it against histories

script/gather_stake_accounts.rb
```diff
@@ -24,6 +24,20 @@
 
   TotalRewardsUpdateService.new(p.payload[:network], p.payload[:stake_accounts_active]).call
 
+  next unless network == 'mainnet'
```
kbiala, line 27: This would skip the logging that we do at the end of this script. We only want to skip the GatherExplorerStakeAccountsService service.

kbiala (changes requested):

app/services/gather_explorer_stake_accounts_service.rb
```diff
@@ -0,0 +1,80 @@
...
+
+    @logger.info("total results from solana stakes: #{@stake_accounts.count}}")
+  end
+
+  def update_stake_accounts
+    batch_no = 0
+    @stake_accounts.in_groups_of(BATCH_SIZE) do |batch|
+      batch_no += 1
+      updated_stake_accounts = []
+      @logger.info("processing batch #{batch_no} of #{@stake_accounts.count / BATCH_SIZE}")
```
kbiala, line 43:
```
      @logger.info("processing batch #{batch_no} of #{(@stake_accounts.count / BATCH_SIZE.to_f).ceil}")
```

PR #875, opened 2023-10-27
186322522-tweak-epoch-detection

maciejkocylapc:
#### What's this PR do?
- make sure the explorer stake accounts views are never empty

- run tests
- check if there are no empty explorer stake account views after new epoch shows up

- [URL to the PivotalTracker ticket](https://www.pivotaltracker.com/story/show/186322522)

#### Things to watch out for

### DIFF
--- a/app/queries/explorer_stake_account_query.rb
+++ b/app/queries/explorer_stake_account_query.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class ExplorerStakeAccountQuery
-  MIN_ACCOUNTS_NUMBER = Rails.env.test? 1 : 500_000
+  MIN_ACCOUNTS_NUMBER = Rails.env.test? ? 1 : 500_000
 
   def initialize(withdrawer: nil, staker: nil, vote_account: nil, stake_pubkey: nil, network: "mainnet")
     @withdrawer = withdrawer
--- a/app/queries/explorer_stake_account_query.rb
+++ b/app/queries/explorer_stake_account_query.rb
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
 class ExplorerStakeAccountQuery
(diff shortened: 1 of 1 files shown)


app/queries/explorer_stake_account_query.rb
```diff
@@ -12,9 +14,16 @@ def initialize(withdrawer: nil, staker: nil, vote_account: nil, stake_pubkey: ni
   def call(page: 1, per: 20)
     explorer_stake_accounts = ExplorerStakeAccount.where(
       network: @network,
-      epoch: EpochWallClock.by_network(@network).first&.epoch
+      epoch: EpochWallClock.by_network(@network).last&.epoch
```
kamtarnowski, line 17: `EpochWallClock.by_network(@network).last&.epoch` can be wrap into some variable and used in line 23 as well.

app/queries/explorer_stake_account_query.rb
```diff
@@ -12,9 +14,16 @@ def initialize(withdrawer: nil, staker: nil, vote_account: nil, stake_pubkey: ni
...
     )
 
+    if explorer_stake_accounts.count < MIN_ACCOUNTS_NUMBER
```
kamtarnowski, line 20: `EpochWallClock.by_network(@network).last` can be nil? If so then `count` method will raise an error.

kamtarnowski (changes requested):

kamtarnowski:
The tests have failed.

app/queries/explorer_stake_account_query.rb
kbiala, line 17: Why changed to .last? 
Epochs are in descending order in by_network scope

Also, we should only query .count here, to avoid pulling the collection twice, for two different epochs.

kbiala (approved):

PR #879, opened 2023-11-06
186256273 geofence ips by country

maciejkocylapc:
#### What's this PR do?
- determine the country of origin of the user
- restrict staking links on validators detail page for a specific country

- run tests
- uncomment last line in set_geo_country method and try various IPs
- check if links are displayed according to condition

- [URL to the PivotalTracker ticket](https://www.pivotaltracker.com/story/show/186256273)

#### Things to watch out for


app/controllers/concerns/ip_geo_country_helper.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module IpGeoCountryHelper
+  def set_geo_country
+    db = MaxMindDB.new("./GeoLite2-Country.mmdb")
+    db.lookup(request.remote_ip).country.iso_code
```
kbiala, line 6: In Load Balancer, this will return private IP, not original request IP.
Check X-Forwarded-For, X-Forwarded-Proto, and X-Forwarded-Port headers that DO load balancer sets to pass information about original request. And if they're empty, we'll need to change load balancer configuration.

kbiala: https://docs.digitalocean.com/products/kubernetes/how-to/configure-load-balancers/#preserving-client-source-ip-address

PR #885, opened 2023-11-23
186442305 save validator images

maciejkocylapc:
#### What's this PR do?
- save validator avatars as image files and display it instead of avatar uls

- install ImageMagick
- run bundle install
- run migrations
- run tests
- start sidekiq
- get avatar_urls by `rails r script/validators_get_info.rb` (cli) and `rails r script/validators_get_avatar_url.rb` (keybase)
- download files for existing urls by  `rails r script/update_validator_avatars.rb`
- check logs at log/update_avatar_file_service.log
- check if avatars are attached to validators with avatar_url ( i.e. validator.avatar.attached? )
- check if avatars displayed for validators on pages like validators index, details etc are grabbed from storage

2nd Scenario:
- Change avatar_url link for one of the validators to test new links detection
- Confirm that the job was created
- After the job is completed, confirm that the url was updated and file downloaded and attached

3rd scenario:
- Test files presence in API responses

- [URL to the PivotalTracker ticket](https://www.pivotaltracker.com/story/show/186442305)

#### Things to watch out for


app/services/update_avatar_file_service.rb
```diff
@@ -0,0 +1,83 @@
...
+    Digest::MD5.file(@tmp_file).hexdigest
+  end
+
+  def process_and_save_avatar
+    @avatar_file = STORAGE_PATH + "/" + @validator.avatar_file_name
+    ImageProcessing::MiniMagick
+      .source(@tmp_file)
+      .convert("png")
+      .resize_to_limit(*IMAGE_SIZE_LIMIT)
+      .call(destination: @avatar_file)
```
kbiala, line 59:
```suggestion
    ImageProcessing::MiniMagick.source(@tmp_file)
                               .convert("png")
                               .resize_to_limit(*IMAGE_SIZE_LIMIT)
                               .call(destination: @avatar_file)
```

app/javascript/packs/mixins/validators_mixins.js
```diff
@@ -39,7 +39,9 @@ Vue.mixin({
     },
 
     avatar_url(validator) {
-      if(validator['avatar_url']) {
+      if(validator['avatar_file_url']) {
+        return validator['avatar_file_url']
+      } else if (validator['avatar_url']) {
```
kbiala, line 44: I think the whole point of this story was to not display original URL addresses. So we should not display avatar_url, only avatar_file_url or default image.

script/update_validator_avatars.rb
```diff
@@ -0,0 +1,19 @@
+#frozen_string_literal: true
+
+require "digest/md5"
+require "open-uri"
+
+TMP_STORAGE_PATH = Rails.root.join("storage", "validators", "avatars", "tmp").to_s.freeze
```
kbiala, line 6: Is this constant needed here? There's already path definition inside service.

test/services/update_avatar_file_service_test.rb
```diff
@@ -0,0 +1,61 @@
+#frozen_string_literal: true
+require "test_helper"
+
+class UpdateAvatarFileServiceTest < ActiveSupport::TestCase
+  include ActiveJob::TestHelper
+  include VcrHelper
+
```
kbiala, line 7: Redundant new line

kbiala (changes requested):
Configuration for aws is missing.
For production and staging, we should update from `config.active_storage.service = :local` to `config.active_storage.service = :amazon`

app/helpers/validators_helper.rb
```diff
@@ -99,7 +99,11 @@ def create_avatar_link(validator)
                     content_tag(:span, nil, class: "fa-solid fa-users-slash", title: "Private Validator"),
                     class: "img-circle-large-private"
       end
-    elsif validator.avatar_url.present?
+    elsif validator.avatar.attached?
+      link_to validator_url(link_params) do
+        image_tag validator.avatar, class: "img-circle-large"
+      end
+    elsif validator.avatar_url.present? && false
```
kbiala, line 106: Same comment here. We don't want to display images from original URLs.

kbiala (changes requested):
I think we should also remove avatar_url from FIELDS_FOR_API

app/models/validator.rb
```diff
@@ -145,6 +152,14 @@ def total_active_stake
     end
   end
 
+  def avatar_file_url
+    polymorphic_url(avatar) if avatar.attached?
+  end
+
+  def update_avatar_file
+    UpdateAvatarFileWorker.perform_async({validator_id: id}.stringify_keys) unless avatar_url.nil?
```
kbiala, line 160: Let's add this worker to "low_priority" queue instead of default one.

kbiala (changes requested):

app/workers/update_avatar_file_worker.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class UpdateAvatarFileWorker
+  include Sidekiq::Worker
```
kbiala, line 4: I'd also add retry limit.

kbiala (changes requested):

script/update_validator_avatars.rb
```diff
@@ -0,0 +1,19 @@
...
+
+NETWORKS.each do |network|
+  # next unless network == 'mainnet'
+
+  Validator.where(network: network).where.not(avatar_url: nil).each do |validator|
```
kbiala, line 11: Is there a point in processing inactive validators? I'd only process active ones.

app/javascript/packs/mixins/validators_mixins.js
maciejkocylapc: true, but the images won't download instantly, so I thought we could leave it for the transition period

kbiala: Makes sense, I'll create a follow-up story to remove them

script/update_validator_avatars.rb
```diff
@@ -0,0 +1,15 @@
...
+
+require "digest/md5"
+require "open-uri"
+
+NETWORKS.each do |network|
+  Validator.where(network: network, is_active: true).where.not(avatar_url: nil).each do |validator|
+    begin
+      UpdateAvatarFileService.new(validator).call
+    rescue
+      puts "Error updating avatar for #{validator.account} on #{validator.network}"
```
kbiala, line 11: This log is not really useful.
We should at least log the error, if not send it to Appsignal.

app/services/update_avatar_file_service.rb
```diff
@@ -0,0 +1,82 @@
...
+  end
+
+  def download_tmp_file
+    download = URI.open(@validator.avatar_url)
+    @tmp_file = STORAGE_PATH + "/tmp/" + @validator.avatar_tmp_file_name
+    if IO.copy_stream(download, @tmp_file).positive?
+      @logger.info("Downloaded file: " + @tmp_file)
+      @tmp_file
+    else
+      @logger.error("Error downloading file: " + tmp_file_path)
```
kbiala, line 44: There's no such variable as tmp_file_path

app/services/update_avatar_file_service.rb
```diff
@@ -0,0 +1,82 @@
+#frozen_string_literal: true
+
+require "digest/md5"
+require "open-uri"
+require "image_processing/mini_magick"
+
+STORAGE_PATH = Rails.root.join("storage", "validators", "avatars").to_s.freeze
```
kbiala, line 7: This path look a little messed up. There's a different path defined in config/storage, and on localhost the files are saved in root/storage, not in root/storage/validators/avatars.
There're some temporary files stored in storage/validators/avatars and storage/validators/avatars/tmp - before and after resizing, converting etc - but since both versions are temporary, why dividing to tmp and not tmp, or create so nested directories just for temporary files?

kbiala (changes requested):

app/services/update_avatar_file_service.rb
```diff
@@ -0,0 +1,82 @@
...
+  end
+
+  def tmp_file_md5
+    Digest::MD5.file(@tmp_file).hexdigest
+  end
+
+  def process_and_save_avatar
+    @avatar_file = STORAGE_PATH + "/" + @validator.avatar_file_name
+    ImageProcessing::MiniMagick.source(@tmp_file)
+                               .convert("png")
```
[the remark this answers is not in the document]
kbiala: There's still a problem with animated gifs. I think we should allow them, so they should not be converted to png.

PR #888, opened 2023-11-30
186547814 update validators api

maciejkocylapc:
#### What's this PR do?
- set random order of the results by score
- remove inactive validators from default query (make it optional)
- add is_active field to API response
- update docs

- check if the docs are updated
- run tests
- try to send requests to /api/v1/validators/:network with various params and see if they match the documentation description
- make sure there are no inactive validators by default

- [URL to the PivotalTracker ticket](https://www.pivotaltracker.com/story/show/186547814)

#### Things to watch out for

### DIFF
--- a/app/queries/validator_query.rb
+++ b/app/queries/validator_query.rb
@@ -12,7 +12,19 @@ class ValidatorQuery < ApplicationQuery
                      end
   end
 
-  def call(network: "mainnet", sort_order: "score", limit: 9999, page: 1, random_seed_val: 123, query_params: { admin_warning: nil, query: nil, jito: false })
+  def call(
+      network: "mainnet",
+      sort_order: "score",
+      limit: 9999,
+      page: 1,
+      random_seed_val: 123,
+      query_params: {
+        admin_warning: nil,
+        query: nil,
+        jito: false,
+        active_only: true
+      }
+    )
     scope = @default_scope.preload(:validator_score_v1_for_api)
     scope = filter_by_collaboration(scope, query_params[:jito])
     scope = filter_by_network(scope, network)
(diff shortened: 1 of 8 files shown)


kbiala (approved):

app/queries/validator_query.rb
```diff
@@ -21,7 +33,7 @@ def call(network: "mainnet", sort_order: "score", limit: 9999, page: 1, random_s
     scope = set_ordering(scope, sort_order, random_seed_val)
     scope = set_pagination(scope, page, limit)
 
-    @api ? scope : scope.scorable
+    query_params[:active_only] ? scope.scorable : scope
```
kbiala, line 36: This condition is incorrect. All strings, even those with "false" values, are truthy, so this condition will always return active validators.
Please add this scenario to tests - only the opposite scanario is tested.

app/views/public/api_documentation.html.erb
```diff
@@ -118,8 +120,8 @@
     "skipped_slot_percent":"0.5155",
     "ping_time":"205.703",
     "url":"<%= @api_path %>/validators/testnet/12345j84ecQEKUvQ7gYMKxKwKF6PbYXxXxxXXXXXxXXX.json",
-    "stake_pools_list":["Jito", "Marinade"]
-
+    "stake_pools_list":["Jito", "Marinade"],
+    "is_active": true
```
kbiala, line 124: I can't see "is_active" attribute in response, no matter which request params I set.

kbiala (changes requested):

test/queries/validator_query_test.rb
```diff
@@ -254,4 +237,30 @@ class ValidatorQueryTest < ActiveSupport::TestCase
 
     assert_equal validator, result.first
   end
+
+  test "#call returns only active validators if active_only is true" do
```
kbiala, line 241: Please also test the result if active_only is empty.

kbiala (approved):

PR #904, opened 2023-12-15
186442321 - add sitemap

kamtarnowski:
#### What's this PR do?
- Applies sitemap and robots.txt

- instructions designed both for development and staging platforms
- run `rake sitemap:refresh:no_ping` in order to create sitemaps. The output should include information about compressed sitemap(s) files
- run server and enter /robots.txt. Verify disallowed paths (pages where the robot should not process).
- download the sitemap file over the path /sitemap.xml.gz. It will include an information about indexed sitemaps (sitemap1.xml.gz, sitemap2.xml.gz and so on). Download those files and verify whether they have links to the application sites (validators, data_centers, vote_accounts, explorer_stake_accounts and others mentioned in config/sitemap.rb file).

[- URL to the PivotalTracker ticket](https://www.pivotaltracker.com/story/show/186442321)

#### Things to watch out for


config/sitemap.rb
```diff
@@ -0,0 +1,99 @@
...
+                                           "http://www.example.com"
+                                         end
+
+sitemap_url = if Rails.env.production?
+                "https://www.validators.app/sitemap.xml.gz"
+              elsif Rails.env.stage?
+                "https://stage.validators.app/sitemap.xml.gz"
+              else
+                "http://www.example.com/sitemap.xml.gz"
+              end
```
kbiala, line 15: You can use variable here, to reduce to one condition:
```
host = if Rails.env.production?
         "https://www.validators.app"
         ...
         ...
       end
SitemapGenerator::Sitemap.default_host = host
sitemap_url = "#{host}/sitemap.xml.gz"
```

config/robots/robots.stage.txt
```diff
@@ -0,0 +1,4 @@
+User-Agent: *
+Disallow: /
+
+Sitemap: https://stage.validators.app/sitemap.xml.gz
```
kbiala, line 4: I think we should block the same pages as we block for production

kbiala (changes requested):

kbiala (approved):

PR #967, opened 2024-02-21
187087318 - skip network if stake pools don't exist

kamtarnowski:
#### What's this PR do?
- As in the title.

- run `rails script/gather_stake_accounts.rb` and check if the pipeline is run only for the network where stake pools exist.

[- URL to the PivotalTracker ticket](https://www.pivotaltracker.com/story/show/187087318)

#### Things to watch out for


script/gather_stake_accounts.rb
```diff
@@ -5,6 +5,8 @@
 
 include StakeLogic
 NETWORKS.each do |network|
+  next unless StakePool.where(network: network).exists?
```
kbiala, line 8: I would only add some info like "Skipping #{network}, because there are no StakePools" or similar
Otherwise looks good

PR #1063, opened 2024-12-20
188704452-tweak-ping-times

maciejkocylapc:
#### What's this PR do?
- add max limit in API ping times request
- add validation to collector payload
- update docs
- remove unused code 

- run tests
- try both collector and ping_times endpoints
- try to provide incomplete payload to collector and make sure the data does not get saved

- [URL to the PivotalTracker ticket](https://www.pivotaltracker.com/story/show/188704452)

#### Things to watch out for

### DIFF
--- a/app/logic/collector_logic.rb
+++ b/app/logic/collector_logic.rb
@@ -3,6 +3,7 @@
 # CollectorLogic
 module CollectorLogic
   include PipelineLogic
+  REQUIRED_PAYLOAD_FIELDS = %w[network avg_ms min_ms max_ms from_account to_account from_ip to_ip].freeze
 
   # guard_ping_times will check the collector payload type & version
   #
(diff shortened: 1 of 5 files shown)


app/logic/collector_logic.rb
```diff
@@ -23,6 +24,10 @@ def ping_times_guard
         return Pipeline.new(400, p[:payload], 'Wrong payload_version')
       end
 
+      REQUIRED_PAYLOAD_FIELDS.each do |field|
+        return Pipeline.new(400, p[:payload], "No payload field: #{field}") unless collector.payload =~ /#{field}/
+      end
```
kbiala, line 29: It generally works, but we should check the same for every record in the pipeline (now it's enough that one of them has required fields).

kbiala (changes requested):

PR #1102, opened 2025-07-16
add hidden flag to policy

maciejkocylapc:
#### What's this PR do?
- add hidden flag and do not display hidden policies on index page

- run migration
- set policys hidden to true
- make sure it's not displayed on index page

#### Things to watch out for

### DIFF
--- a/app/models/policy.rb
+++ b/app/models/policy.rb
@@ -6,6 +6,7 @@
 #  additional_metadata :json
 #  description         :text(65535)
 #  executable          :boolean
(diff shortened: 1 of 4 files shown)


app/models/policy.rb
```diff
@@ -61,4 +62,8 @@ def to_builder
   def non_validators
     policy_identities.where(validator_id: nil)
   end
+
+  def self.visible
```
kbiala, line 66: this should be a scope

PR #1110, opened 2025-08-12
sc 166657 use mtr command instead of traceroute

maciejkocylapc:
#### What's this PR do?

- install mtr (via brew)
- on mac follow these instructions https://blog.dave-bell.co.uk/2020/01/06/using-mtr-on-os-x-without-sudo/
- get some ips for validators via `rails r script/data_centers_scripts/append_data_centers_geo_data.rb`
- test scripts e.g. `rails r script/data_centers_scripts/fix_data_centers_ovh.rb`

#### Things to watch out for

### DIFF
--- a/config/schedule.rb
+++ b/config/schedule.rb
@@ -69,15 +69,15 @@ every 1.hour, at: 35, roles: [:background] do
   ruby_script_data_centers 'assign_data_center_scores.rb'
 end
 
-every 1.hour, at: 40, roles: [:background] do
+every 1.day, at: '6:00am', roles: [:background] do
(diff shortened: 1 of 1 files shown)


kbiala (changes requested):
It doesn't work, because returned data format is different.
In old version it returned string, it new version - it returns json. 
When you try to use this json in scripts, it returns an error, because match method doesn't exist for json (eg. https://github.com/brianlong/www.validators.app/blob/master/script/data_centers_scripts/fix_data_centers_reliablesite.rb#L50).
Refactoring should not change returned data types, get_matching_traceroute method should return string.

kbiala (approved):

PR #1123, opened 2025-09-17
sc-168521-exclude-11111111111111111111111111111111

maciejkocylapc:
#### What's this PR do?
- hide non-validator identities that were replaced by 1s

- run tests
- see the policy detail view - there should be no identities with only 1's listed

#### Things to watch out for

### DIFF
--- a/app/models/policy.rb
+++ b/app/models/policy.rb
@@ -66,6 +66,6 @@ class Policy < ApplicationRecord
   end
 
   def non_validators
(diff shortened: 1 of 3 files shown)


kbiala (changes requested):
identities_count doesn't exclude invisible indentities (it's displayed on the index page, and also in API)

kbiala (approved):

